PR #19, opened 2017-04-27
Add Fedora and CentOS installation instructions

xvitaly:
Purple-discord package is now available at Fedora's main repository. Added Fedora and CentOS installation instructions.


EionRobb:
Nice one. Thanks again @xvitaly :)

Issue #25, opened 2017-04-29
Channel and role mentions


Eloston:
I tried using `@username#id` as well as omitting certain parts but they all don't work. Doing `<@id>` made the web client stylize that text like a username mention.

EionRobb:
There's no way at the moment.

We might borrow the same idea that the [bitlbee-discord](https://github.com/sm00th/bitlbee-discord) plugin [does](https://github.com/sm00th/bitlbee-discord/blob/88de238b8317f9f19da6d12d786a66004a3f81e5/README#L105-L114) and try to see if someone typed "@EionRobb" or "EionRobb:" in order to mention them

Eloston:
I'm not familiar with bitlbee-discord. What happens if there are two users with the same name but different IDs?

Eloston:
Okay, so I found out after some experimentation that using `<@USER_ID>` where `USER_ID` is the ID of the user will properly notify the user. But this is probably some Discord feature I didn't know about before since it also works on the web client.

Eloston:
So I looked into this a little bit earlier and I've come up with an idea. It involves overriding the tab-completion signal in Pidgin:

* Username mentioning: Replace strings of the format `@unicode_string#discriminator` with a mention. Tab completion logic is a bit complicated, though:
    * For regular ASCII-only usernames, behave similarly to Pidgin's default autocomplete behavior, except that it only functions if there's a `@` behind the cursor. That is, start matching from the beginning of the username and then show options if there are multiple usernames sharing the same prefix.
    * For unicode characters, ignore any characters that don't have an "ASCII-equivalent". Also, convert unicode characters to ASCII-looking equivalents.
        * But if there are already non-ASCII characters between the cursor and the last `@`, then follow the same behavior employed by ASCII-only usernames.
        * After a match has been found in either mode, replace the characters with the full unicode username.
    * If the cursor immediately follows a `@`, autocomplete the last person that mentioned us. If such person does not exist, then autocomplete the last person that sent a message. (I'm not entirely sure if this feature is a good idea, though)
    * If #14 gets implemented, then we want to be able to tab-complete either nicknames or usernames.
* Channel mentions: Replace strings of the format `#unicode_string`.
    * Autocompletions will be limited to channels existing within the currrent guild (unless we find some way of specifying a guild).
    * ASCII and Unicode behavior will work identically to what usernames have.

EionRobb:
[...]

PR #31, opened 2017-05-13
Improve channel mentions

Eloston:
Improve channel mentions by including the guild name and using `discord_normalise_room_name`. If the guild name isn't available for some reason, just have `#` followed by the channel name.

This also fixes a crash when the client receives a channel mention with an unknown channel id.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -1280,10 +1280,10 @@ discord_replace_channel(const GMatchInfo *match, GString *result, gpointer user_
 		//TODO make this a clickable link
 		guild = discord_get_guild_int(da, channel->guild_id);
 		if (guild) {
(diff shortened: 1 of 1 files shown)


EionRobb:
Nice :) Thanks :+1:

mimi89999:
So how do I mention?

Eloston:
@mimi89999 There's no "user-friendly" way of instantiating any mention right now. The only way right now is to use [Discord's extensions to the Markdown format](https://discordapp.com/developers/docs/resources/channel#message-formatting) when composing your message.

mimi89999:
So I need to right click on user, info, get the ID and type `<@ID>`? Not quite intuitive...

EionRobb:
@mimi89999 it's still a work in progress. See https://github.com/EionRobb/purple-discord/issues/25

PR #38, opened 2017-05-21
Add FPM make targets for building packages.

doctorlard:
This requires the rather nifty [FPM](https://github.com/jordansissel/fpm) tool, which for simple projects like this can build RPMs, Debian, and others quite easily.


EionRobb:
What are the resulting files named?  Should the Makefile target be those names instead of the 'phoney' deb/rpm?

doctorlard:
Ah well maybe, but we've got the versions in there. Also, the version thing is using hg to get the hash and revision number, whereas in git land it should use something like
GIT_COMMIT = $(shell git log -1 --pretty='format:%h')
It's also not including the pixmaps. Maybe I'll do some more work...!

doctorlard:
I updated the commit to include the pixmaps. I think keeping them as PHONY targets is better because the target file is dynamically named from versions etc.

EionRobb:
I'm still only seeing the one commit in the PR, not another one that adds the deb/rpm to PHONY, nor includes the pixmaps?

xvitaly:
@doctorlard Never use such tools as fpm, alien, etc. They build bad, malformed packages.

You must use only native tools to build packages for each platform. Native SPEC for RPM package can be found [here](https://github.com/xvitaly/purple-discord).

bobbybee:
This PR seems to be stale, closing.

PR #47, opened 2017-06-01
Wrapped error to support libpurple3 build

Fank:
And it works

image

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -94,9 +94,9 @@ json_object_to_string(JsonObject *obj)
 #include <purple.h>
 #if PURPLE_VERSION_CHECK(3, 0, 0)
 #include <http.h>
-#define purple_notify_error_wrap(a,b,c,d) purple_notify_error(a,b,c,d,NULL)
+#define purple_notify_error_wrap(a,b,c,d,e) purple_notify_error(a,b,c,d,e)
 #else
-#define purple_notify_error_wrap(a,b,c,d) purple_notify_error(a,b,c,d)
+#define purple_notify_error_wrap(a,b,c,d,e) purple_notify_error(a,b,c,d)
 #endif
 
 #ifndef PURPLE_PLUGINS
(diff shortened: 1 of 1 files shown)


Fank:
@EionRobb better?

EionRobb:
Apologies, I didn't explain myself very well :) 

I mean, to not use a wrapper but instead to use the purple_notify_error() function in the code and then add a fallback for purple2

The intention is to target the code to purple3 primarily, with compat code, and only ever so slight amounts of purple3 compat code where its completely unavoidable

Fank:
I know what you mean but i don't know how to overwrite a function call with a preprocessor macro.
The only solution i know is to use a wrapper macro.

Fank:
I think i may found a solution, but could not test it, only have a pidgin 3 enviroment.

Fank:
@EionRobb i think this is a good solution, if it works :D

EionRobb:
I was just looking at what the 5th parameter is supposed to be in purple3 for `purple_notify_error()` and it looks like it should be the callback from `purple_request_cpar_from_connection(pc)`, but we already expand that out to `#define purple_request_cpar_from_connection(a)  purple_connection_get_account(a), NULL, NULL` for the `purple_request_input()` calls. 

So the line should be
`purple_notify_error(da, _("Bad channel type"), _("Cannot join a voice channel as text"), "", purple_request_cpar_from_connection(da->pc));`
and then that would expand to the equivalent of
`purple_notify_error(da, _("Bad channel type"), _("Cannot join a voice channel as text"), "", da->account, NULL, NULL);` so that certainly complicates the `purple_notify_error()` call a lot more.  Hummm.

What a can of worms you have opened 😄

Fank:
Di you want to take a look in it, you are more familiar with it, than me?
Or should i try to resolve it?

EionRobb:
Lets get this merged so that it'll work on purple3, and I'll tidy up for purple2 :)

Thanks for your work, and thanks for talking it through

Fank:
@EionRobb I send you a mail yesterday, i would be thankful if you could answer it. :D

PR #48, opened 2017-06-03
Implement custom emojis

bobbybee:
Users can add custom emojis on a per-server basis, which are encoded
with a special in-text notation like mentions. This patch enables
sending these emojis through the expected :syntax: as well as displaying
them as links to the image. Future revisions may display the image
itself, although this is not currently implemented due to Pidgin/purple
quirks.

This functionality is separate from the standard emojis provided in
Discord, which (confusingly) use the same syntax. This emojis, like
:stuck_out_tongue: are sent directly as Unicode, and are therefore
handled correctly out-of-the-box, provided an appropriate smiley theme
is installed. Again, even without the patch, these can be sent with
Pidgin's emoji selector. Future revisions may unify the syntax like in
the Discord client, but this functionality requires the inclusion of an
emoji dictionary.

Emojis can only be sent from the current selected server, like in the
normal client without purchasing Discord Nitro. It is unknown if these
limitations are enforced server-side or not; either way, changing this
may be legally risky.

Signed-off-by: Alyssa Rosenzweig <alyssa@rosenzweig.io>

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -278,6 +278,8 @@ purple_message_destroy(PurpleMessage *message)
 #define IGNORE_PRINTS
 
 static GRegex *channel_mentions_regex = NULL;
+static GRegex *emoji_regex = NULL;
+static GRegex *emoji_natural_regex = NULL;
 
 typedef enum{
 	USER_ONLINE,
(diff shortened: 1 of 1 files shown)


EionRobb:
Nice one, thanks :)

PR #50, opened 2017-06-03
Translate actions between /me acts and _acts_

bobbybee:
From the users perspective, both Discord and Pidgin correctly denote
actions by sending "/me eats" or similar. However, while Pidgin
works with this format internally, Discord does not render commands like
"/me eats" as anything but that string; instead, it erronously converts
the action in non-semantic markdown denoted with underscores "_eats_".
Without the patch, actions can still be sent between Purple users, but
interoperability with normal Discord users breaks. This patch applies a
regex in each direction to convert between Purple's (superior) notation
and Discord's native format. The only known edge case is if a Discord
user manually types out the underscore syntax for emphasis, but
anecdotally this is exceedingly rare.

Signed-off-by: Alyssa Rosenzweig <alyssa@rosenzweig.io>

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -1,6 +1,7 @@
 /*
  *   Discord plugin for libpurple
  *   Copyright (C) 2016  Eion Robb
+ *   Copyright (C) 2017 Alyssa Rosenzweig
  *
  *   This program is free software: you can redistribute it and/or modify
  *   it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


EionRobb:
Might need to use `purple_message_meify()` instead of a regex to detect /me when there's formatting applied

EionRobb:
Nice one.  Thanks again :)

PR #52, opened 2017-06-07
Fetch direct messages on connect

bobbybee:
This enables direct messaging to work as expected when the user is offline; this also means that for many users, it will not be necessary to use discordapp.com at all to check messages.

Additionally, the same code gets the channels that the user is mentioned in; however, fetching channel history is somewhat more involved and therefore not covered in this patch set. It is currently stubbed out, but prints out the name of the channel with mentions for debugging.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -2302,10 +2302,15 @@ discord_got_guilds(DiscordAccount *da, JsonNode *node, gpointer user_data)
 	discord_socket_write_json(da, obj);
 }
 
+/* If count is explicitly specified, use a static request (DMs).
+ * If it is not, use a dynamic request (rooms).
+ * TODO: Possible edge case if there are over 100 incoming DMs?
+ */
+
 static void
 discord_get_history(DiscordAccount *da, const gchar *channel, const gchar *last, int count)
 {
-	gchar *url = g_strdup_printf("https://" DISCORD_API_SERVER "/api/v6/channels/%s/messages?limit=%d&around=%s", channel, count ? count : 100, last);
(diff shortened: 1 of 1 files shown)


libdiscord.c
```diff
@@ -2298,6 +2302,40 @@ discord_got_guilds(DiscordAccount *da, JsonNode *node, gpointer user_data)
...
+		guint mentions = json_object_get_int_member(state, "mention_count");
+
+		if(mentions) {
+			gboolean isDM = g_hash_table_contains(da->one_to_ones, channel);
+
+			if(isDM) {
+				discord_get_history(da, channel, last_id, mentions * 2);
+			} else {
+				/* TODO: fetch channel history */
+				printf("%d unhandled mentions in channel %s\n", mentions, discord_get_channel_global(da, channel)->name);
```
EionRobb: Can you switch out the `printf()` for a `purple_debug_misc()` instead, otherwise this can mess up the formatting of Finch

libdiscord.c
```diff
@@ -2298,6 +2302,40 @@ discord_got_guilds(DiscordAccount *da, JsonNode *node, gpointer user_data)
 	discord_socket_write_json(da, obj);
 }
 
+static void
+discord_get_history(DiscordAccount *da, const gchar *channel, const gchar *last, const int count)
+{
+	gchar *url = g_strdup_printf("https://" DISCORD_API_SERVER "/api/v6/channels/%s/messages?limit=%d&around=%s", channel, count ? count : 100, last);
```
EionRobb: Does `discord_get_history()` need to be called multiple times if there's more than 100 messages to fetch?

EionRobb: Is "around" for messages before and after the timestamp?  Should it be before or after?

EionRobb:
rude

PR #56, opened 2017-06-08
Misc codebase cleanup

bobbybee:
This has a number of misc codebase cleanups, which will (hopefully) improve maintainability and minimise work needed for code audits etc:

* Move the purple3 compatibility code into its own file, which should probably be its own repository anyway.
* Remove a great deal of dead code, including some unused debug functions which are no longer useful but add complexity to the codebase
* Pass through an automatic styling system (astyle), to ensure a uniform code style (tabs, K&R braces..)
* Run the codebase through two static analysis tools (cppcheck and clang) and fix minor resulting issues (nothing serious found -- congrats!)
* Remove unnecessary hg command which causes an ignorable error on git platforms
* Flip typenames to a single line for consistency (I don't care one way or the other, but I had to pick one and this was somewhat easier for the regexs to handle.)


p4plus2:
Some of these are pretty good, however the debug functions are still very much useful.  They are disabled by default as they can cause a *lot* of lag in channels with many users (One of the test servers I use has over 10k users.)  So, they are there as a compile time option, though the config to enable and disable could be made better.  Also that switch is still important, there are other types we don't support yet, so, thats not one to keep either.  As a different question why did you move the json wrappers with the compatibility wrappers?  That is not a very logical location IMO.  Pretty sure "discord_mark_conv_seen" shouldn't have been removed either...  Overall I think a few of these are ready to merge and others probably not so much.  I'll wait for what @EionRobb  has to say though.

bobbybee:
Thanks for the feedback (I wasn't sure with a lot of the changes). I'll resubmit a PR with cherrypicked changes when I have a chance.

PR #74, opened 2017-08-11
Escape markdown correctly

### LINKED ISSUE #55: Issues with markdown escapes break emojis, links, etc
This regression is the result of the naive markdown escape I implemented. A proper refactor would help tremendously to solve quirks like this.

bobbybee:
Rather than escaping all markdown special characters, this routine scans the string character-by-character, only inserting backslashes conservatively. This closes #55


r4m0n:
Confirming URLs with '_' working perfectly now :-)

bobbybee:
Hooray! :tada:

PR #75, opened 2017-08-11
Fix direct message ack'ing

### LINKED ISSUE #57: Direct Messages repeat
Commit 4c4c480 introduced a change that causes direct messages to be fetched from the discord server multiple times. Looks like a fix would entail either setting those messages as read on the discord server or not refetching them.

bobbybee:
This PR closes #57 

Note: this has not been extensively battletested due to lack of ~~guinea pigs~~ online users. But.. it compiles? :wink:

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -394,6 +394,7 @@ typedef struct {
 
 	GHashTable *one_to_ones;      // A store of known room_id's -> username's
 	GHashTable *one_to_ones_rev;  // A store of known usernames's -> room_id's
+	GHashTable *last_message_id_dm; // A store of known room_id's -> last_message_id's
 	GHashTable *sent_message_ids; // A store of message id's that we generated from this instance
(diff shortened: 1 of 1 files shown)


libdiscord.c
```diff
@@ -2196,6 +2205,7 @@ discord_build_groups_from_blist(DiscordAccount *ya)
 			discord_id = purple_blist_node_get_string(node, "discord_id");
 			if (discord_id != NULL) {
 				g_hash_table_replace(ya->one_to_ones, g_strdup(discord_id), g_strdup(name));
+				g_hash_table_replace(ya->last_message_id_dm, g_strdup(discord_id), "0");
```
EionRobb: Don't put a `const char *` (the `"0"`) into a hash table when you've specified the value destroy function is `g_free` (in your `g_hash_table_new_full` init)

PR #77, opened 2017-08-12
Buddy list autopopulation

### LINKED ISSUE #45: Able to see and join but silent failure sending to locked channels

### LINKED ISSUE #53: Channel ID shown instead of Name in Pidgin 2 Buddy List
Just as the title suggests. I'm running Pidgin 2.12 from Debian Stretch.

I don't remember exactly what commit caused this, but I think it started happening late May, so around the time of this commit: ec1292a00dcc2ae833ba527c951574234a1519e3

bobbybee:
* Automatically populate the buddy list with each group representing a server
* Implement permission computation routines
* Prefer this interface which is saner to work with (this closes #53)
* Hide rooms without permission to READ_MESSAGES (this closes #45)
* Hide voice channels

The room list API is deprecated but still included; bitlbee, spectrum2, etc require it.

The major flaw of this code is that it can only create the buddy list groups, not update them. We'll have to figure out a scheme for synchronisation later. As a workaround for the time being, deleting the group and reloading the plugin will sync it.

Another question to consider is handle user aliases, removed chats, etc. Is it possible to disable these changes in the UI?

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,10 +34,6 @@ Requires devel headers/libs for libpurple and libjson-glib [libglib2.0-dev, libj
 	sudo make install
 ```
 
-Where's my rooms?
------------------
-In Pidgin, look in Tools->Room List to get a list of rooms you're in.  It's recommended that you add them to your buddy list and mark them as 'persistent' and 'auto-join'.
-
 Show your appreciation
 ----------------------
 Did this plugin make your life happier?  [Send me $1](https://www.paypal.com/cgi-bin/webscr?cmd=_s-xclick&hosted_button_id=PZMBF2QVF69GA) to say thanks!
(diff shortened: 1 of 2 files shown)


EionRobb:
I'm wondering if it should be a setting to autocreate them?
Does this create rooms that have been muted?

bobbybee:
Probably (for both questions).

EionRobb:
> Probably (for both questions).

So should that be added to this PR before or after merging?

bobbybee:
Shouldn't matter, you decide? 😆

bobbybee:
It's now an optional setting. Muted channels are still visible -- this can be revisited once we implement muting from Pidgin. (For the time being, respecting this setting is nontrivial for little benefit.)

Issue #82, opened 2017-08-12
Rate limiting


bobbybee:
Although nothing we do is automated, to stay on the safe side we need to implement rate limiting as per the docs.

LurkerHub:
6 years later, people are still getting temp-banned for exceeding rate limits.  Can this please be prioritized?

Issue #24, opened 2017-04-29
Add unread messages to channel window when joining


Eloston:
Having this for when the user joins a channel that has been joined before and for joining new channels would be nice.

Maybe an option to configure how many of such messages to retrieve would be nice to prevent overloading the user (and potentially the client). A harcoded amount of 100 wouldn't be too bad either.

EionRobb:
Issue #4 will also be fixed when we fix this

bobbybee:
A naive solution is implemented for this, but it seems to cause more problems than not at the moment.

alexolog:
What is the status of this issue?

dlmarquis:
This was fixed in commit 1ec4d36944380258847557d37026bdfcb9a4915f.

PR #90, opened 2017-08-15
Disable recursive room history fetching until rate limiting is implemented

bobbybee:
See #82 and #24 .... the plugin is annoying (and potentially risky) to use in large active channels. Until we can sort out the "correct" way to handle history -- which requires modifying libpurple and/or hacking around the spec with a custom Pidgin plugin -- it's safest to disable it altogether.


EionRobb:
Nah

LurkerHub:
Issue still happens

PR #99, opened 2017-08-19
Extra mentions

### LINKED ISSUE #25: Channel and role mentions

bobbybee:
* Send channel mentions
* Send role mentions
* Send everyone mentions
* Beep on role mentions
* Colour role mentions correctly

This closes #25 

No tab completion but hey

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -1362,6 +1362,46 @@ discord_replace_mentions_bare(DiscordAccount *da, DiscordGuild *g, gchar *messag
 	return message;
 }
 
+static guint64
+discord_find_role_by_name(DiscordGuild *guild, gchar *name)
+{
+	if (purple_strequal(name, "everyone")) {
+		return guild->id;
+	}
+
+	GHashTableIter iter;
+	gpointer key;
+	gpointer value;
+	g_hash_table_iter_init(&iter, guild->roles);
+
+	while (g_hash_table_iter_next(&iter, (gpointer *) &key, &value)) {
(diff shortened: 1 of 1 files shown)


libdiscord.c
```diff
@@ -1351,17 +1386,76 @@ discord_replace_mention(const GMatchInfo *match, GString *result, gpointer user_
...
+			g_free(message);
+			message = tmp;
+		}
 	}
 
 	return message;
 }
 
+static guint64
+discord_find_role_by_name(DiscordGuild *guild, gchar *name)
```
EionRobb: In the future, use `const gchar *name` so that the compiler/person calling the function knows that the value isn't going to be modified

libdiscord.c
```diff
@@ -1611,20 +1720,30 @@ discord_process_message(DiscordAccount *da, JsonObject *data)
...
+					flags |= PURPLE_MESSAGE_NICK;
+					break;
+				}
+			}
+		}
 	}
 
-	/* Replace <#channel_id> with channel names */
-	tmp = g_regex_replace_eval(channel_mentions_regex, escaped_content, -1, 0, 0, discord_replace_channel, da, NULL);
+	escaped_content = discord_replace_mentions_bare(da, guild, escaped_content);
```
EionRobb: Does this mean that every message that comes through is then processed for user/channel mentions, even if there are none?

PR #100, opened 2017-08-19
Channel IDs instead of names; closes #39

### LINKED ISSUE #39: Can't join 2 channels on 2 different servers with the same name at the same time.
So I tried opening a channel called #general on two servers in pidgin but it seems that once one instance of said channel is open pidgin associates all channels with the same name as one in the same despite them being on different servers. Double clicking on a channel with the same name as an already joined channel on another server will not join the channel on the other server but will just bring the chat window to the front and display the channel with that name that was connected to first.

bobbybee:


libdiscord.c
```diff
@@ -3813,16 +3813,7 @@ discord_open_chat(DiscordAccount *da, guint64 id, gchar *name, gboolean present)
...
-		chatconv = purple_conversations_find_chat_with_account(name, da->account);
-	}
-
-	if (chatconv == NULL) {
-		/* TODO FIXME */
-		gchar *chat_name = g_strdup_printf("%" G_GUINT64_FORMAT, id);
-		chatconv = purple_conversations_find_chat_with_account(chat_name, da->account);
-		g_free(chat_name);
-	}
+	chatconv = purple_conversations_find_chat_with_account(from_int(id), da->account);
```
EionRobb: `from_int()` allocates memory so this line will leak without a `g_free()`

libdiscord.c
```diff
@@ -3832,7 +3823,7 @@ discord_open_chat(DiscordAccount *da, guint64 id, gchar *name, gboolean present)
 		return NULL;
 	}
 
-	chatconv = purple_serv_got_joined_chat(da->pc, g_int64_hash(&id), name ? name : g_strdup_printf("%" G_GUINT64_FORMAT, id));
+	chatconv = purple_serv_got_joined_chat(da->pc, g_int64_hash(&id), from_int(id));
```
EionRobb: And this one (although it looked like it leaked before, too)

bobbybee:
Fixed

PR #103, opened 2017-08-19
Use Discord read_state in lieu of tracking ourselves

bobbybee:
If this works (unclear), this makes read state behaviour much more predictable by using Discord's acking mechanism instead of rolling our own

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -168,6 +168,7 @@ typedef struct {
 	GHashTable *one_to_ones;		/* A store of known room_id's -> username's */
 	GHashTable *one_to_ones_rev;	/* A store of known usernames's -> room_id's */
 	GHashTable *last_message_id_dm; /* A store of known room_id's -> last_message_id's */
+	GHashTable *read_state;			/* A store of known room_id's -> read_state */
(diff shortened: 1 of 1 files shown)


bobbybee:
(Ready for review :tada:)

EionRobb:
So we're no longer going to fill in history if we've seen it on another device?

bobbybee:
No, although to be fair, it's not like that worked correctly to begin with :smile:

alexolog:
Aside: It would be great if history handling with multiple devices could be made consistent across protocols.  
As it stands, XMPP does not repeat messages that were seen on other devices while Hangouts and SkypeWeb do.  Eion, your thoughts?

bobbybee:
(Unsolicited thoughts follow)

One answer is to run libpurple on a server ala Quassel(-core). That is,
the connection XMPP/Discord/Hangouts/Skype would be from the
centralised, always-on machine. The various devices would be thin
clients only. This centralised server can remember which device saw
which messages, rather than having the prpls manage it themselves. 

Additionally, this architecture gives a natural place to implement a web
client... think WebPidginZ but better? :-)

PR #104, opened 2017-08-19
Fix emoji/mentions, closes #87

### LINKED ISSUE #87: Custom emojis don't work on the same line as a mention (?)

bobbybee:


EionRobb:
Magically free'ing and modifying a function parameter is not a reusable pattern (eg, it won't allow you to pass in const strings).  This needs to go back to returning a newly allocated gchar* and the calling function can free the message as appropriate

EionRobb:
Obsoleted by 9e2369e2bc6540692ff5c7b7c7710fdf7c228630

PR #106, opened 2017-08-20
Roles in info

### LINKED ISSUE #23: Add user role info
I have two ideas relating to user roles that would be nice to have:

1. For users with special roles such as Owner, Moderator, Bot, etc., place special icons next to their name and sort them above others in the member list
2. Add a method of retrieving the list of all user roles. ~Perhaps using the dialog used to get more user info will work.~ EDIT2: The "Buddy Information" dialog box doesn't work for this.

EDIT: This isn't related to the general topic of this issue, but adding availability status (Online, Idle, Do Not Disturb) and the game status to the more info dialog box would be nice too.

bobbybee:
Closes #23 as well as makes the pane prettier ❤️

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -1260,6 +1260,9 @@ discord_replace_channel(const GMatchInfo *match, GString *result, gpointer user_
 	return FALSE;
 }
 
+#define COLOR_START "<font color=\"#%06X\">"
+#define COLOR_END "</font>"
+
 static gboolean
 discord_replace_role(const GMatchInfo *match, GString *result, gpointer user_data)
 {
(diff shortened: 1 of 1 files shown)


libdiscord.c
```diff
@@ -4470,8 +4472,9 @@ discord_got_info(DiscordAccount *da, JsonNode *node, gpointer user_data)
 		purple_notify_user_info_add_pair_html(user_info, _("In-Game"), user->game);
 	}
 
-	purple_notify_user_info_add_pair_html(user_info, NULL, NULL);
-	purple_notify_user_info_add_pair_html(user_info, _("Connected Accounts"), NULL);
```
EionRobb: Can we keep the headings?

bobbybee:
Restored (although I thought it was pretty obvious in context...)

PR #109, opened 2017-08-20
Translation

### LINKED ISSUE #60: Translation support
Not all strings that should be translated are marked as translatable (with the `_()` glib/gettext function).  Make sure all the strings are marked and then use that to provide a .pot file to translate

bobbybee:
* Poorly integrate with gettext (this closes #60 )
* Add a poorly written Spanish translation to demo


Makefile
```diff
@@ -26,6 +26,9 @@ LDFLAGS ?= -Wl,-z,relro
 
 CFLAGS  += -std=c99 -DDISCORD_PLUGIN_VERSION='"$(PLUGIN_VERSION)"'
 
+# CFLAGS += -DLOCALEDIR=\"$(PIDGIN_DATADIR)/locale\"
+CFLAGS += -DLOCALEDIR=\"/usr/share/locale/\"
```
EionRobb: use  ```-DLOCALEDIR="`$(PKG_CONFIG) --variable=datadir purple`/locale"```  instead

libdiscord.c
```diff
@@ -25,6 +25,24 @@
...
+#include <purple.h>
+#if PURPLE_VERSION_CHECK(3, 0, 0)
+#include <http.h>
+#endif
+
+#ifndef PURPLE_PLUGINS
+#	define PURPLE_PLUGINS
+#endif
+
+#define ENABLE_NLS 1
```
EionRobb: Would this be better as a compile-time option?

libdiscord.c
```diff
@@ -2843,8 +2854,8 @@ discord_login_response(DiscordAccount *da, JsonNode *node, gpointer user_data)
 								 _("Enter Discord auth code"),
 								 _("You can get this token from your two-factor authentication mobile app."),
 								 NULL, FALSE, FALSE, "",
-								 _("_Login"), G_CALLBACK(discord_mfa_text_entry),
-								 _("_Cancel"), G_CALLBACK(discord_mfa_cancel),
+								 _("Login"), G_CALLBACK(discord_mfa_text_entry),
+								 _("Cancel"), G_CALLBACK(discord_mfa_cancel),
```
EionRobb: The `_` before the text is for the accelerator key which is important for screen-readers, eg Alt+L on windows will trigger the login

po/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+xgettext ../libdiscord.c -k_ --no-location -o purple-discord.pot
```
EionRobb: It would probably be better to have .pot generation done in the Makefile

po/es.po
```diff
@@ -0,0 +1,175 @@
...
+msgstr "Entrar un servidor..."
+
+msgid "nick <new nickname>:  Changes nickname on a server"
+msgstr "nick <nombre nuevo>:  Cambiar el nombre en un servidor"
+
+msgid "kick <username>:  Remove someone from channel"
+msgstr "kick <nombre>:  Quitar alguien del canal"
+
+msgid "leave:  Leave the channel"
+msgstr "salir: Salir el canal"
```
EionRobb: The command is "leave" not "salir"

po/es.po
```diff
@@ -0,0 +1,175 @@
...
+
+msgid "part:  Leave the channel"
+msgstr "salir: Salir el canal"
```
EionRobb: The command is "part" not "salir"

po/compile.sh
```diff
@@ -0,0 +1 @@
+msgfmt es.po -o es.mo
```
EionRobb: This should probably be in the Makefile too

bobbybee:
The above issues have been resolved.

Makefile
```diff
@@ -26,13 +26,18 @@ LDFLAGS ?= -Wl,-z,relro
 
 CFLAGS  += -std=c99 -DDISCORD_PLUGIN_VERSION='"$(PLUGIN_VERSION)"'
 
+LOCALEDIR = $(shell $(PKG_CONFIG) --variable=datadir purple)/locale
```
EionRobb: This needs to be inside the not-windows block, since the LOCALEDIR on windows is "$(PROGRAMFILES)/Pidgin/locale" rather than anything set with pkg-config

libdiscord.c
```diff
@@ -4576,7 +4582,7 @@ discord_status_text(PurpleBuddy *buddy)
 			return NULL;
 		}
 
-		return g_markup_printf_escaped(_("Playing %s"), user->game);
+		return g_markup_printf_escaped("%s %s", _("Playing"), user->game);
```
EionRobb: I don't think we can assume the order here, especially for languages that are 'backwards' compared to Latin-based languages (Japanese comes to mind)

libdiscord.c
```diff
@@ -4547,16 +4553,16 @@ discord_status_types(PurpleAccount *account)
...
+	status = purple_status_type_new_with_attrs(PURPLE_STATUS_AWAY, "idle", _("Idle"), TRUE, use_status_as_game, FALSE, "message", _("In-Game"), purple_value_new(PURPLE_TYPE_STRING), NULL);
 	types = g_list_append(types, status);
 
-	status = purple_status_type_new_with_attrs(PURPLE_STATUS_UNAVAILABLE, "dnd", _("Do Not Disturb"), TRUE, use_status_as_game, FALSE, "message", "In-Game", purple_value_new(PURPLE_TYPE_STRING), NULL);
+	status = purple_status_type_new_with_attrs(PURPLE_STATUS_UNAVAILABLE, "dnd", _("Do Not Disturb"), TRUE, use_status_as_game, FALSE, "message", _("In-Game"), purple_value_new(PURPLE_TYPE_STRING), NULL);
 	types = g_list_append(types, status);
 
-	status = purple_status_type_new_with_attrs(PURPLE_STATUS_OFFLINE, "offline", _("Offline"), TRUE, FALSE, FALSE, "message", "In-Game", purple_value_new(PURPLE_TYPE_STRING), NULL);
+	status = purple_status_type_new_with_attrs(PURPLE_STATUS_OFFLINE, "offline", _("Offline"), TRUE, FALSE, FALSE, "message", _("In-Game"), purple_value_new(PURPLE_TYPE_STRING), NULL);
```
EionRobb: Would it be more simple to translate if we used "Playing" here as well instead of "In-Game"?

Makefile
```diff
@@ -26,13 +26,16 @@ LDFLAGS ?= -Wl,-z,relro
...
 CFLAGS  += -std=c99 -DDISCORD_PLUGIN_VERSION='"$(PLUGIN_VERSION)"'
 
+# Comment out to disable localisation
+CFLAGS += -DENABLE_NLS
+
 # Do some nasty OS and purple version detection
 ifeq ($(OS),Windows_NT)
   DISCORD_TARGET = libdiscord.dll
   DISCORD_DEST = "$(PROGRAMFILES)/Pidgin/plugins"
   DISCORD_ICONS_DEST = "$(PROGRAMFILES)/Pidgin/pixmaps/pidgin/protocols"
```
EionRobb: So close!  We just need LOCALEDIR added for Windows, then it's all good to go.

Did you want to add that as part of this PR or want me to add after?

bobbybee:
Either way. I suppose you doing it is better, since Windows ;P

PR #110, opened 2017-08-20
Match fingerprint with user agent

bobbybee:


EionRobb:
You and I (and @p4plus2 too!) have talked a few times about this about whether it's better to broadcast the fact that we're using Pidgin, or try to hide and pretend to be an official client.

I'm still not 100% sure what's "best".  Should we be pretending to be an official client, but say we're running on the OS that we actually are?

bobbybee:
I don't know, but regardless, we should match with the useragent I
think.

Issue #80, opened 2017-08-12
Can't Compile on Linux Mint 17.3 (Trusty 14.04) - Mercurial hg?


WithThisHerring:
Following instructions at https://github.com/EionRobb/purple-discord#compiling , I enter
```
$ git clone git://github.com/EionRobb/purple-discord.git
Cloning into 'purple-discord'...
remote: Counting objects: 662, done.
remote: Compressing objects: 100% (116/116), done.
remote: Total 662 (delta 133), reused 179 (delta 98), pack-reused 446
Receiving objects: 100% (662/662), 290.32 KiB | 0 bytes/s, done.
Resolving deltas: 100% (417/417), done.
Checking connectivity... done.
$ cd purple-discord
$ make
```
and the following appears:
```
abort: there is no Mercurial repository here (.hg not found)
cc -fPIC -O2 -g -pipe -Wall -std=c99 -DDISCORD_PLUGIN_VERSION='"0.9.2017.08.12"' -shared -o libdiscord.so libdiscord.c -Wl,-z,relro  `pkg-config purple glib-2.0 json-glib-1.0 --libs --cflags`   -Ipurple2compat -g -ggdb
```
Optimistically hoping that error wouldn't matter and I could skip to checkinstall... (Checkinstall per [xvitaly's advice](https://github.com/EionRobb/skype4pidgin/issues/488#issuecomment-254660940) on my question regarding your Skype plugin :)  )
`$ sudo checkinstall` 
With result (super-long, so stuck in pastebin):
https://pastebin.com/ScugQiKB

I have googled the error, but I can't find anything that is relevant to this plugin or my OS (Linux Mint 17.3, based on Ubuntu Trusty 14.04).  Mercurial is installed.  Pidgin 2.10.9 (libpurple 2.10.9) is installed.  I don't think the advice on [this closed issue applies](https://github.com/EionRobb/purple-discord/issues/64), as eeboaz uses very different installation commands, but I could be horribly wrong.  I am still ignorant of too much Linux stuff, so please be very explicit in any help you give me.  Thank you!

bobbybee:
It looks like it did compile?

WithThisHerring:
Some of the last lines from trying checkinstall are
```
Building Debian package... FAILED!
 
*** Failed to build the package
```
so I didn't think so?  Something went wrong...

[...]

Issue #64, opened 2017-07-03
Unable to build on FreeBSD - Makefile requires mercurial but clone is not under mercurial repo


eeboaz:
```
$ gmake
gmake: hg: Command not found
echo "You need libpurple development headers installed to be able to compile this plugin"
You need libpurple development headers installed to be able to compile this plugin
```

I installed mercurial via pkg and ran gmake again:
```
$ gmake
abort: there is no Mercurial repository here (.hg not found)
```

Can I just change these out with the appropriate git calls?

EionRobb:
hg is just optional.  The important part is about needing the libpurple development headers.  What's the output of `pkg-config --libs --cflags purple-2` ?

eeboaz:
I think the verbiage of the mercurial message (abort) was causing me to focus on the wrong issue.

pkg-config (freedesktop.org) isn't available for FreeBSD. I'll add some FreeBSD specific updates to the Makefile (within ifeq blocks) and see if I can get it to build. I'll send a PR once I get it working.


FYI, I do have the libpurple installed - they're in /usr/local/include/libpurple on FreeBSD:

```
$ pkg query %n-%v | grep purple
libpurple-2.12.0_2
```

EionRobb:
what about if you use  `PKG_CONFIG=pkgconf gmake`?

eeboaz:
This definitely helped -- I was going about this the wrong way.  No changes are needed for the current Makefile. I was able to build this for FreeBSD using your suggestions and studying a makefile from another (similar) FreeBSD port to determine what I was missing.

The plugin is working great with Pidgin.  I'll work with one of the FreeBSD ports committers to get this submitted to the FreeBSD ports collection.

Thank you very much for your assistance!

bobbybee:
Closing as the issue seems to be fixed :smile:

PR #114, opened 2017-08-21
Remove hg specific code from Makefile

bobbybee:
#64, #80, ... it confuses people for functionality which I don't think is used by anyone but Eion 🙄

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,13 +13,7 @@ DIR_PERM = 0755
 LIB_PERM = 0755
 FILE_PERM = 0644
 
-REVISION_ID = $(shell hg id -i)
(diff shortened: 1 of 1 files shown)


EionRobb:
How about using `git` with an `hg` fallback, so that we have the best of both worlds?

bobbybee:
I'm not convinced this is functionality we need in the Makefile anyway..

EionRobb:
It's there to know what version of the plugin was built, in the absence of real version numbers

xvitaly:
@EionRobb Please merge this pull request.

EionRobb:
Without a git fallback (or a git+hg fallback) we don't have a reasonable way of knowing what dev version of the plugin people are using

PR #117, opened 2017-08-22
Italian Translation

Penaz91:
I added an Italian *.po to the po/ directory, hopefully I didn't do anything wrong

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -1,0 +1,174 @@
+#
+# Penaz <penazarea@altervista.org>, 2017.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: 0.9\n"
+"Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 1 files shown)


EionRobb:
Grazie mille 🙂

bobbybee:
Unless the latest batch of changes fixed this, there also has to be a
corresponding change in the makefiles. I'll fix this when I get home in
a few hours.

Additionally, I think you forgot to translate "You can get this token
from your two-factor authentication mobile app.", unless those all
happen to be Spanish cognates :p

Penaz91:
Definitely not "Spanish cognates", I'm fixing it.

I don't know how I managed to miss that line xD

I'm adding a line to install the italian translation in the makefile, under the "it" folder, although I'm not sure if it's the right directory, since "it_IT" would fit too.

I'm making a pull request for both

PR #118, opened 2017-08-22
Fixed Italian Translation

Penaz91:
As @bobbybee pointed out, there was a missing line in the translation, as well as no install line in the makefile, this should fix those both.

I'm really sorry for disturbing again, I have no idea how I managed to miss it.

Thanks for your patience.


bobbybee:
Cool, thanks :-)

EionRobb:
I think we needed to change how we process the `install-locales` command instead of just accepting this as-is.  I guess it can go in a separate commit, but we should be updating the Makefile to compile any .po in the po folder, rather than specify each language individually

bobbybee:
Yeah, I'm working on it (and a few other related issues)

PR #133, opened 2017-10-09
Fix bitlbee segfault

esainane:
On bitlbee, discord_roomlist_serialize is called via
prplcb_roomlist_get_room_name via prplcb_roomlist_add_room, which crashes
as the fields are null. We add guild_name to the fields for the category,
which means guild_name gets used as the serialized name for categories.


bobbybee:
(Will test this later this evening)

EionRobb:
Thanks @esainane  ! 👍

PR #134, opened 2017-10-17
Flag changes for Mac

SIELA1915:
Changed LDFLAGS for Mac (-z option doesn't exist) and install -D (option doesn't exist on mac either)


bobbybee:
Can someone with a mac test this?

PR #146, opened 2017-11-12
Fallback to git when constructing PLUGIN_VERSION

kevinoid:
The Makefile attempted to use the output of `hg id` for constructing `PLUGIN_VERSION`.  This fails since the project is not using Mercurial.  Presumably this was copied from another project, because it has been present since the initial commit.

This PR uses the git short commit hash instead.  Since git doesn't provide an equivalent to the Mercurial local revision number, that part is removed.

Note:  Git errors are suppressed since the project may be built on systems without git installed (e.g. distro build systems/chroots) or without the git metadata (e.g. released sources or downloaded
zip/tarball from GitHub).

Thanks for considering,
Kevin

P.S.  You may want to consider using the most recent commit date instead of the current date to make the build reproducible (e.g. `git show -s --format=%cd.%h --date=format:%Y.%m.%d`).  If so, I can update this PR to include that change as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,10 +13,14 @@ DIR_PERM = 0755
 LIB_PERM = 0755
 FILE_PERM = 0644
 
-REVISION_ID = $(shell hg id -i)
-REVISION_NUMBER = $(shell hg id -n)
+# Note: Use "-C .git" to avoid ascending to parent dirs if .git not present
+GIT_REVISION_ID = $(shell git -C .git rev-parse --short HEAD 2>/dev/null)
+REVISION_ID = $(shell hg id -i 2>/dev/null)
+REVISION_NUMBER = $(shell hg id -n 2>/dev/null)
 ifneq ($(REVISION_ID),)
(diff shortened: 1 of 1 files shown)


bobbybee:
> The Makefile attempted to use the output of `hg id` for constructing `PLUGIN_VERSION`.  This fails since the project is not using Mercurial.  Presumably this was copied from another project, because it has been present since the initial commit.

Eion uses Mercurial with this project, although I suppose nobody else
does.

kevinoid:
@bobbybee Ah, that makes sense.  Thanks!  I could change the PR to just silence the `hg id` errors if that would be preferable.

EionRobb:
I do use hg, yeah, but I wouldn't complain about a PR that tries to use either git or hg (or falls back to current date, if neither)

kevinoid:
@EionRobb Sounds good.  I changed the PR to add a fallback to `git` and to silence errors from both `git` and `hg`, since errors are expected to occur normally.  Let me know what you think.

EionRobb:
Nice one, thanks! :)

kevinoid:
My pleasure.  Thanks for the quick review and merge!

bobbybee:
It's his strategy to indoctrine new contributors.

It works :)

PR #165, opened 2017-12-27
Preserve order of guild groups

### LINKED ISSUE #163: Pidgin groups keep moving to the top
Whenever the account is enabled and reconnected, or connected for the first time, the default groups created for each server (guild?) keep moving to the top of the buddy list, even after I've moved them around.

tpenguinltg:
Fixes #163

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -2856,7 +2856,7 @@ discord_buddy_guild(DiscordAccount *da, DiscordGuild *guild)
 		discord_add_channel_to_blist(da, channel, group);
 	}
 
-	purple_blist_add_group(group, NULL);
+	purple_blist_add_group(group, group->node.prev);
 }
 
 static void


bobbybee:
Nicely done.

EionRobb:
This isn't purple3 compatible. Structs are opaque and functions should be users to get the previous group instead

bobbybee:
....oops?

EionRobb:
The cause of the problem is that `purple_blist_add_group` shouldn't be called if the group is already in the blist, otherwise the group will get moved around

tpenguinltg:
So if the call to `purple_blist_find_group` earlier in the function returns something, can it be assumed that it's already in the blist?

EionRobb:
@tpenguinltg yes, that's correct. If the add-group function was moved inside the "if not exists" check then it would only add if needed :)

PR #166, opened 2017-12-27
Only add group to blist when creating a new one

### LINKED ISSUE #163: Pidgin groups keep moving to the top

tpenguinltg:
Preserves guild group order in a way that's compatible with purple3.

Fixes #163

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -2818,10 +2818,12 @@ discord_buddy_guild(DiscordAccount *da, DiscordGuild *guild)
 
 	if (!group) {
 		group = purple_group_new(guild->name);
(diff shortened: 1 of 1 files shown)


tpenguinltg:
Take two. Hopefully this is better than #165.

EionRobb:
That is superlative :)

Thanks for taking the time to fix it up 👍

bobbybee:
:tada:

*apologising to Eion intensifies*

PR #188, opened 2018-06-01
Do not discard obtained user name

andreygursky:
Fixes regression since 322bad567a504d67a41939c18916a442cff9e17b.


EionRobb:
Whoops, nice catch, thanks!

PR #202, opened 2018-10-19
Add how to install Libpurple

Nic-Wow1:


EionRobb:
The packages needed for compiling are in the first line of the 'Compiling' section already: "Requires devel headers/libs for libpurple and libjson-glib [libglib2.0-dev, libjson-glib-dev and libpurple-dev]"

bobbybee:
Per Eion's comment, closing

PR #204, opened 2018-10-19
Add Make install Instructions

Nic-Wow1:


bobbybee:
make is part of the standard build-essential.

PR #209, opened 2018-11-23
Fetch own avatar

bobbybee:
Both Discord and libpurple maintain the concept of account icons, like buddy icons but for the account holder herself. Unfortunately, these do not get much use in prpls (even libpurple native prpls), because they are scarcely used in the Pidgin UI. The Discord UI, by contrast, uses "avatars" everywhere, so it makes sense to have this available.

This patch modifies get_avatar/got_avatar to support self-avatars, as well as hooks in this function at READY.

It further manages checksums itself via purple_account settings, as checksumming is normally only available for buddy icons.

It can be confirmed working by navigating to the account settings pane in Pidgin, as well as playing a much more active role in mythical avatar-heavy libpurple-based clients.


bobbybee:
Thank you!

PR #218, opened 2018-12-08
Alt logo

bobbybee:
Replacing the logo with a simple, aesthetically-similar but non-copyrightable design.


pabs3:
If the existing logos are copied from the proprietary codebase, they are probably under a license that does not allow redistribution, so it was a bad idea to copy them into the codebase, especially without indicating in the README that they were actually not original works created independently.

It is likely that Discord doesn't give out trademark licenses or if it does, different ones are given independently to each requester. This means that using the logo to refer to something other than the official client is likely to be a violation of their trademarks. I guess the name purple-discord might also be a violation, but that is less clear to me and would need clarification from a lawyer.

The combination of these two things means I probably have to remove purple-discord from Debian unless PR is merged soon.

In this case the PNG versions of the images are clearly no longer "preferred form for modification", since they are automatically generated from the SVG file, so they should not be present in the source code repository, please remove the PNG versions of the images and depend on ImageMagick or rsvg or another SVG renderer. If you refuse to do this, there *must* be a mechanism that cleans up pre-built stuff so that distros can ensure that they can build from source, I suggest that `make clean` also delete those files. If the files get removed and are thus always generated, please mention them in the .gitignore file too.

bobbybee:
> Just wondering, has Discord actually made an issue of this?

No, but it is an issue for distros, and we're trying to play it safe.

> it was never problem

Remember gaim?

szr8:
On 12/17/2018 15:02 PM, Alyssa Rosenzweig wrote:
> > Just wondering, has Discord actually made an issue of this?
>
> No, but it is an issue for distros, and we're trying to play it safe.
I can understand that, though I don't see how Linux distros (or other 
operating systems for that matter) can be held responsible for the 
content of packages coming from third parties.

> > it was never problem
>
> Remember gaim?
It's been a while, but I don't remember anything specific there in 
regards to icons or similar, and a search on the web came up empty. What 
exactly happened?

pabs3:
Debian is responsible to both copyright holders and to users for the
things it distributes, including purple-discord.

https://www.debian.org/social_contract

In a similar way, Microsoft's GitHub platform is also responsible for
the things they redistribute, but they have the advantage of the DMCA
safe harbour provisions that Debian does not.

Wikipedia has the details about the Pidgin/Gaim legal dispute:

https://en.wikipedia.org/wiki/Pidgin_(software)#Naming_dispute

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

EionRobb:
Just to clarify some of the history around the icons - the logos were based on creative commons licensed artwork found via Google image search (with the CC license filter option), then modified by myself to use the closest blue from the Tango Icon Theme Guidelines colour palette in order to better fit the Tango guidelines.  Whether the source (not sure if I know which one it specifically is now, would have to hunt through browser history) that I based my changes from had the right to license what could be considered a derivative artwork I'm not sure, but its definitely murky enough (well, slightly more murky than the legal grey-area we fit into by writing a 3rd party client for a closed protocol) to consider replacing the likeness of the character that discord uses in its logo with a generic and most likely un-copywritable letter D.  I can't find any registered trademark for the discord logo however, so there's not a trademark infringement, just a maybe-possibly copyright infringement, depending on how your jurisdiction perceives the recreation of an icon.  But just to definitively point out that at no point did I extract any material out of any proprietary client or website belonging to Discord to use as the basis for the icons in this plugin.

With regards to this PR:
Converting from svg isn't ideal for the smaller icons, in terms of following the tango icon guidelines (kinda makes things blurry) and kinda breaks the windows builds.  That said, I have no strong objections to this PR if you could make some kind of response about the scaling and windows build points, @bobbybee

bobbybee:
> ish

I don't see how the svg in question could have possibly been CC-licenced
if it's a clear derivative work of a proprietary image? They might've
slapped the CC license on it, doesn't make it kosher.

pabs3:
It could have been independently created, it is a rather simple logo after all.
Not sure if that still makes it a derivative work though.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

bobbybee:
https://commons.wikimedia.org/wiki/Commons:Fan_art#Re-drawing_does_not_avoid_copyright_infringement

pabs3:
Interesting, thanks for the reference.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

szr8:
Why exactly does Microsoft get safe harbor provisions but not Linux 
distributions like Debian?

PR #225, opened 2018-12-19
Updated Italian Translation

Penaz91:
I updated the po file for the Italian language to include the new strings. Sorry for being late!


bobbybee:
Grazie! ;)

PR #226, opened 2018-12-26
Don't animate avatars (specify png)

bobbybee:
See comment for why this is a good idea and shouldn't regress anyone (https://xkcd.com/1172/)


bobbybee:
Ack, committed po's as well, sorry. Take 2.

PR #232, opened 2019-01-17
Strip timestamps when building PNG files

pabs3:
Otherwise the build process will not be deterministic.

See-also: https://reproducible-builds.org/
See-also: https://wiki.debian.org/ReproducibleBuilds/TimestampsInPNG


EionRobb:
I like both stripping and reproducing.  Two thumbs up from me.

PR #236, opened 2019-02-07
Fix off-by-one on history retrieval

### LINKED ISSUE #159: Off-by-one in history loading
When opening a room from the buddy list, all but the most recent message get loaded into the scrollback. This behaves consistently, i.e. if I leave Pidgin offline for a while and open it later, the previous missing message gets filled in at the top of other new ones, and opening/closing the chat repeatedly doesn't cause the last message to appear.

meinwald:
Original code for discord_got_history_of_room breaks before the last message has been processed. This reverses the order of the discord_process_message call and the break and should fix #159 . Note that I'm not particularly familiar with the code so I don't necessarily know all the implications of this change, but it does seem to work in my personal testing.


EionRobb:
Good catch :)

Thanks for this!

PR #242, opened 2019-04-27
Update fr.po

draelenax:
New translation and correct few mistakes

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -8,7 +8,7 @@ msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2018-12-16 11:00-0800\n"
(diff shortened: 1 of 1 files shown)


po/fr.po
```diff
@@ -182,19 +182,19 @@ msgid "Join a server..."
...
 msgid "leave:  Leave the channel"
 msgstr "leave: Quitter le salon"
 
 msgid "part:  Leave the channel"
-msgstr "part: partir du salon"
+msgstr "part: Partir du salon"
 
 msgid "pinned:  Display pinned messages"
-msgstr ""
+msgstr "Afficher les messages épinglés"
```
EionRobb: Should this one have "pinned: " at the start of it to indicate the /slash command name?

draelenax: you're right.
I'm correcting that.

EionRobb:
Fantastic thank you. :)

PR #253, opened 2019-08-03
Add message for requiring CAPTCHA

Eloston:
Adds an error message for when a CAPTCHA is required, and recommends users to use Harmony to authenticate.


EionRobb:
That'll be handy. Thanks :)

EionRobb:
We could also steal the idea from the bitlbee-discord plugin to use the auth token from the browser

Eloston:
Hmm, that'd be nifty. Right now I have to login to a browser first, then login to Pidgin after the browser's in (Harmony doesn't seem to be working properly... CAPTCHA passes, but two-factor auth doesn't seem to work)

freddii:
CAPTCHA in Harmony is broken. Is there another solution?

PR #256, opened 2019-08-24
Channel categories

### LINKED ISSUE #175: Handling channel categories
Channel categories are shown as a chat in the buddy list, just like channels.
This means I can open a channel category in pidgin and send a message to it, but it does nothing on the server side, so the message just vanishes.

Either remove the categories from the buddy list or turn them into groups on the buddy list.

bobbybee:
Closes #175

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -2896,9 +2896,8 @@ discord_roomlist_got_list(DiscordAccount *da, DiscordGuild *guild, gpointer user
 		PurpleRoomlistRoom *local_category =
 			discord_get_room_category(da, id_to_category, channel->category_id, roomlist, category);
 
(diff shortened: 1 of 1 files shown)


bobbybee:
I also added support for exposing these categories in the Room List view.

bobbybee:
Added some more stuff while I was in the room list implementation so now hidden channels are properly hidden in the room list (which should save some embarassment - most people don't realize hidden channels still leak metadata..)

bobbybee:
Threw in some more roomlist improvements while I was working on that set of code. It should be much happier now -- a shame Pidgin doesn't make much use of it, but alas.

EionRobb:
Classy as always :)

tyvm

PR #257, opened 2019-08-24
Respect notifications settings from web client

### LINKED ISSUE #156: Disable/mute notifications for a channel
In the web interface, I can turn off notifications and mute a channel.

I found no way to do it in Pidgin.  Some channels on a server hold no interest for me, but they keep popping up when somebody mentions `@everybody`.

Is there a way to block those notifications?

notifications

bobbybee:
Closes #156

This MR parses the notifications settings sent in the READY packet (they are undocumented officially; see https://github.com/discordjs/discord.js/pull/1365/files for reference).

It then uses the settings to modify message receive behaviour, allowing channels to be muted, mentions to be suppressed, and @everyone mentions to be suppressed.

The next step will be to allow setting these levels from the libpurple client itself; I'm not sure how best to expose this functionality. After that, I'd like us to think about dropping the large-channel hack, but we'll see where this goes.

This MR needs much more extensive testing, but it compiles and I'm about to go to lunch.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -5531,6 +5531,77 @@ discord_status_types(PurpleAccount *account)
 	return types;
 }
 
+/* If a channel is muted, unmute it, or vice verse */
+
+static void
+discord_toggle_mute(PurpleBlistNode *node, gpointer data)
+{
+	DiscordAccount *da = (DiscordAccount *) data;
+	PurpleChat *chat = PURPLE_CHAT(node);
+
+	DiscordChannel *channel = discord_channel_from_chat(da, chat);
+
+	/* Toggle the mute */
+	channel->muted = !channel->muted;
+
+	/* PATCH /users/@me/guilds/[guild id]/settings
+	 * {"channel_overrides": {"channel_id": {"muted": true}}} */
+
+	DiscordGuild *guild = discord_get_guild(da, channel->guild_id);
+
+	if (guild != NULL) {
+		gchar *channel_id = from_int(channel->id);
+
+		JsonObject *data = json_object_new();
+		JsonObject *override = json_object_new();
+		JsonObject *setting = json_object_new();
+
+		json_object_set_boolean_member(setting, "muted", channel->muted);
+		json_object_set_object_member(override, channel_id, setting);
+		json_object_set_object_member(data, "channel_overrides", override);
+
+		gchar *postdata = json_object_to_string(data);
+
+		gchar *url = g_strdup_printf("https://" DISCORD_API_SERVER "/api/v6/users/@me/guilds/%" G_GUINT64_FORMAT "/settings", guild->id);
+		discord_fetch_url_with_method(da, "PATCH", url, postdata, NULL, NULL);
+
+		g_free(channel_id);
+		g_free(url);
+		g_free(postdata);
+
+		json_object_unref(setting);
+		json_object_unref(override);
+		json_object_unref(data);
+	}
+}
+
+static GList *
+discord_blist_node_menu(PurpleBlistNode *node)
+{
+	/* We only have a menu for chats */
+	if (!PURPLE_IS_CHAT(node))
+		return NULL;
+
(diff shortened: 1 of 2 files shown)


libdiscord.c
```diff
@@ -1747,13 +1758,24 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
 	DiscordGuild *guild = NULL;
 	DiscordChannel *channel = discord_get_channel_global_int_guild(da, channel_id, &guild);
 
+	/* Check if we should receive messages at all and shortcircuit if not */
+
+	if (channel->muted)
```
EionRobb: This should probably also check whether the channel is open in the client, in case the user has intentionally joined a muted channel in order to see messages in pidgin, but not on the web

libdiscord.c
```diff
@@ -3359,6 +3387,65 @@ discord_got_read_states(DiscordAccount *da, JsonNode *node, gpointer user_data)
...
+		/* Apply per-channel overrides */
+		JsonArray *overrides = json_object_get_array_member(settings, "channel_overrides");
+		guint olen = json_array_get_length(overrides);
+
+		for (int j = olen - 1; j >= 0; j--) {
+			JsonObject *override = json_array_get_object_element(overrides, j);
+
+			/* Lookup overriden channel */
+			guint64 channel_id = to_int(json_object_get_string_member(override, "channel_id"));
+			DiscordChannel *chan = g_hash_table_lookup_int64(guild->channels, channel_id);
```
EionRobb: Can we use `channel` instead of `chan` to keep it consistent with other uses?

libdiscord.c
```diff
@@ -3359,6 +3387,65 @@ discord_got_read_states(DiscordAccount *da, JsonNode *node, gpointer user_data)
...
+		/* Apply the guild-global settings */
+		GHashTableIter iter;
+		gpointer key;
+		gpointer value;
+
+		g_hash_table_iter_init(&iter, guild->channels);
+
+		while (g_hash_table_iter_next(&iter, &key, &value)) {
+			DiscordChannel *channel = value;
+			channel->muted = all_mute;
```
EionRobb: Is there an event that mutes/unmutes a channel, while we're connected elsewhere?

EionRobb:
I hope you had a good lunch

bobbybee:
Resolved some of the issues. I also added support for mute/unmuting from Pidgin.

bobbybee:
Resolved everything and fixed a bunch of NULL derefs I introduced.

libdiscord.c
```diff
@@ -2175,6 +2208,19 @@ discord_name_group_dm(DiscordAccount *da, DiscordChannel *channel) {
 	return g_string_free(name, FALSE);
 }
 
+DiscordChannel *
+discord_channel_from_chat(DiscordAccount *da, PurpleChat *chat)
+{
+	/* Grab the ID */
+	GHashTable *components = purple_chat_get_components(chat);
+	const gchar *chat_id = g_hash_table_lookup(components, "id");
```
EionRobb: Potentially, could look at the chat's "name" to get the id.  Can add later though

libdiscord.c
```diff
@@ -5417,6 +5531,77 @@ discord_status_types(PurpleAccount *account)
...
+		gchar *url = g_strdup_printf("https://" DISCORD_API_SERVER "/api/v6/users/@me/guilds/%" G_GUINT64_FORMAT "/settings", guild->id);
+		discord_fetch_url_with_method(da, "PATCH", url, postdata, NULL, NULL);
+
+		g_free(channel_id);
+		g_free(url);
+		g_free(postdata);
+
+		json_object_unref(setting);
+		json_object_unref(override);
+		json_object_unref(data);
```
EionRobb: Does this really need 3 unref's?  Does unreffing the outside not unref the insides too?

EionRobb:
nice nice
nice?
nice!

PR #267, opened 2019-09-14
Drop discriminators for group DM members

bobbybee:
We don't need them, provided it's ambiguous.

That's a big provided. Most everything here is to determine if there's ambiguity.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -1688,6 +1688,15 @@ static gchar *
 discord_create_nickname(DiscordUser *author, DiscordGuild *guild, DiscordChannel *channel)
 {
 	if (!guild) {
+		/* For a group DM, try undiscriminated if unambiguous */
+
+		if (channel && channel->type == CHANNEL_GROUP_DM) {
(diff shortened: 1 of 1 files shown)


EionRobb:
So is this creating an aliased entry into the chat so that we can still right-click->IM on a DM buddy, or does it replace that?

bobbybee:
> So is this creating an aliased entry into the chat so that we can still right-click->IM on a DM buddy, or does it replace that?

Yes, the get_real_name adjustment allows both of that (confirmed working).

Only issue I've noticed is that the colouring of us in the group DM is wrong (it should be coloured special for being us, but it's being coloured like any other nick. I don't know enough about Pidgin's nick colouring to know what the problem is.)

EionRobb:
So the self-colouring isn't working due to a timing issue.  Auto-joined groups won't get it, but manually joined ones will.  Something to do with setting the display name with `purple_chat_conversation_set_nick` but the nick isn't ready to go yet or something

PR #270, opened 2019-10-07
Reduce scope of data variable

pabs3:
Suggested-by: cppcheck -j1 --quiet --enable=all .
[libdiscord.c:4577]: (style) The scope of the variable 'data' can be reduced.


EionRobb:
360noscope gg

PR #273, opened 2019-10-07
Convert URL from http to https

pabs3:


EionRobb:
The "S" is for "super-duper"

Thanks :)

PR #274, opened 2019-11-04
Fix wrong sign in retrieve-history logic

dlmarquis:
Fixes an error from commit 08ea265 where > was used in place of <.

Previously, history retrieval was broken for me with purple2 on Ubuntu18.04. With this change history fetching now works for me locally.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -4913,7 +4913,7 @@ discord_got_history_of_room(DiscordAccount *da, JsonNode *node, gpointer user_da
 		JsonObject *message = json_array_get_object_element(messages, i);
 		guint64 id = to_int(json_object_get_string_member(message, "id"));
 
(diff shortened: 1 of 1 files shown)


EionRobb:
Isn't that saying that "if the current message in the loop is older than the latest message, then display it?"

dlmarquis:
My understanding of the code is that we're iterating through a list of messages sent after the locally-stored last message id, and last_message is what the Discord API reports is the latest message in the channel. In this case we'd want to print messages with ids < last_message to catch up with Discord's servers.

EionRobb:
Yeah, ok, that makes sense.  I keep tripping over the history code and making things worse, so having an outside view on it is much appreciated :)

alexolog:
It doesn't retrieve history for me.  What can I be missing?

PR #292, opened 2020-05-09
Changed most instances of "discordapp.com" to "discord.com"

mikoxyz:
Discord recently changed their URL to "discord.com," and along with that they seemed to have moved most of their stuff over to that domain, including their API (any API requests to "discordapp.com" just redirect to "discord.com"). So, as a measure to prevent breakage if Discord ever decides to remove the redirect, I've replaced most instances of "discordapp.com" with "discord.com." However, since Discord haven't moved over their CDN to "discord.com," I've added a commented line underneath every line that references "cdn.discordapp.com" that has a fixed version with "discord.com."

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -64,7 +64,7 @@
 
 #define DISCORD_BUFFER_DEFAULT_SIZE 40960
 
-#define DISCORD_API_SERVER "discordapp.com"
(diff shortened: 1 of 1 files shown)


EionRobb:
Huh, how about that :)
Pretty significant change on their part


Thanks for the PR :)

It might be worth turning the CDN url into a #define setting too at some point

PR #297, opened 2020-06-24
Set the example make parallelism higher in README.md

Avamander:
Updated the README.md so that it has better defaults, it's 2020 and people don't have only one core CPUs any more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Requires devel headers/libs for libpurple and libjson-glib [libglib2.0-dev, libj
 ```bash
 	git clone git://github.com/EionRobb/purple-discord.git
 	cd purple-discord
-	make
+	make -j4
 	sudo make install
 ```
 


EionRobb:
Was this just a drive-by PR or does it have measurable performance improvements?

Avamander:
I was installing the module, saw the readme, thought to update it. There's really no good reason to default to no parallelism, even if it doesn't currently provide noticeable benefits, if the project grows a bit it will. People constantly forget to update build instructions when it gets to that point, it's a tiny bit annoying.

Avamander:
> The amount of parallelism depends on the number of cores, 4 threads will slow down compilation on a device with only 2 cores, eg older raspberry pi's

First of all, the amount of people on such devices is much much lower than the amount of people on better devices, someone with such a weak device should just lower the amount of parallelism. Secondly, it does not slow things down in certain cases.

> It's better to set your make to use parallelism globally

That has it's own downsides, like making it annoying to override that default should it be required. The ideal solution would be bringing `make` to this century, using CMake is an another possibility.

> Compiling multiple files in parallel rather than all in one thread can be slower

But compiling multiple files sequentially rather than in parallel is absolutely massively slower in a vast majority of cases. Any potential slowdown that _might_ happen in edge case is absolutely not worth the massive slowdown caused by `make`'s obsolete defaults.

> The Makefile in purple-discord uses only a single gcc command (see the above point) so specifying make -j4 is a no-op anyway

As I said, for now. It doesn't really hurt.


Overall, I don't really care if this is changed, I'm used to `make`'s obsolete defaults. :)

PR #300, opened 2020-07-26
Inline image display (rudimentary)

gatlin:
With this patch image attachments will now display inline in text chats. The file is temporarily downloaded to somewhere in `/tmp`, copied into the purple image store, and then removed.

If it detects purple >= 3.0.0 it falls back to the present behavior of simply displaying the URL.

I don't expect this PR to be accepted on the first try; improvements I can think of off the top of my head:

- right now it attempts to treat all attachment URLs as images and falls back to displaying the URL if that fails; I should explicitly check media type.
- I can probably bypass creating a temporary file for this and instead just read the contents of the image directly from the remote resource. But I got this working first!
- Purple 3.0.0 support would obviously be good.
- Private chats

These are things I won't bother with if you aren't interested, but if you are I'll work on them and make any changes you request.


EionRobb:
Can you revert b26055d to keep the whitespace as-is?  Makes the diff un-understandable to just see the code changes

Does this work on Windows?

gatlin:
Ah, thank you! Old habits, etc.

EionRobb:
Thanks for doing that.  Much easier to see what's going on now.    👍 

The general principal looks good.

Some feedback:
* Some of the gio functions being used are too new for winpidgin (which only has glib 2.24)
* `g_file_copy` is a blocking function and will lock up the UI, better to use `g_file_copy_async`, but..
* Doesn't use libpurple's networking stack, so won't use its proxy config, ssl certs, etc - maybe consider using the `discord_fetch_url_with_method` function in the prpl instead?
* Non-Pidgin UI's generally don't show inline images, so would be good to show the original URL regardless 
* Probably shouldn't include translation file changes in the PR :)

gatlin:
Thank you for taking the time. I'll take these into consideration and ping you if / when I have made some updates.

pabs3:
Would it be possible to introduce an option to disable this feature, since some folks will not want images?

PR #329, opened 2021-05-01
Fixes double free when DMing

tsheltonprojects:
purple 2, msg is set, then freed on line 5864, then freed again in discord_created_direct_message_send on the callback, in my case line 5809. which fails/seg faults. tested on my debian/stretch.


EionRobb:
Thanks for that!  I was just about to start digging into why that was happening :)

PR #336, opened 2021-05-06
Inline images, spoiler text, and history off-by-one error

dlmarquis:
Let me know if you want these split out into their own PRs.

I've added `discord_fetch_url`-based image retrieval based somewhat on @gatlin's PR.

Trying to fetch the url from the `url` node gave me an authentication error, so `proxy_url` is used to display the image. It also quietly logs the `url` node because I've previously run into an issue where I couldn't view a logged image because the proxy url expired.

The user has two new account prefs: whether to display images at all, and whether images should also display in large channels. The latter (hopefully) prevents pidgin from choking on a huge number of backlogged images at startup. Spoiler'd images appear as a text link regardless of your settings.

To go along with spoiler'd images, I also added rudimentary support for spoiler text. Spoiler text now appears as black-on-black, so that the user has to highlight text to read it. Sending spoiler text remains as it always has: use Discord's native `||double-bar syntax||`.

The history off-by-one error is an ancient bug I didn't fix from my last PR over a year ago. This should finally fully fix the typo made when that code was refactored.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -5293,7 +5293,7 @@ discord_got_history_of_room(DiscordAccount *da, JsonNode *node, gpointer user_da
 		JsonObject *message = json_array_get_object_element(messages, i);
 		guint64 id = to_int(json_object_get_string_member(message, "id"));
 
-		if (id < last_message) {
+		if (id <= last_message) {
 			rolling_last_message_id = discord_process_message(da, message, DISCORD_MESSAGE_NORMAL);
 		}
 	}
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -234,7 +234,7 @@ typedef struct {
 
 	gboolean compress;
 	z_stream *zstream;
-	
+
 	PurpleHttpKeepalivePool *http_keepalive_pool;
 } DiscordAccount;
 
@@ -379,7 +379,7 @@ discord_new_guild_role(JsonObject *json)
 	guild_role->id = to_int(json_object_get_string_member(json, "id"));
 	guild_role->name = g_strdup(json_object_get_string_member(json, "name"));
 	guild_role->color = json_object_get_int_member(json, "color");
-	
+
 	if (json_object_get_string_member(json, "permissions")) {
 		const gchar *permissions = json_object_get_string_member(json, "permissions");
 		guild_role->permissions = to_int(permissions);
@@ -985,7 +985,7 @@ discord_update_cookies(DiscordAccount *ya, const GList *cookie_headers)
 		const gchar *cookie_end;
 		gchar *cookie_name;
 		gchar *cookie_value;
-		
+
 		cookie_start = cur->data;
 		cookie_end = strchr(cookie_start, '=');
 
@@ -996,7 +996,7 @@ discord_update_cookies(DiscordAccount *ya, const GList *cookie_headers)
 
 			if (cookie_end != NULL) {
 				cookie_value = g_strndup(cookie_start, cookie_end - cookie_start);
-				
+
 				g_hash_table_replace(ya->cookie_table, cookie_name, cookie_value);
 			}
 		}
@@ -1183,7 +1183,7 @@ discord_send_auth(DiscordAccount *da)
 		json_object_set_string_member(properties, "browser_user_agent", DISCORD_USERAGENT);
 		json_object_set_string_member(properties, "browser_version", "51.0.2704.103");
 		json_object_set_string_member(properties, "os_version", "10");
-		
+
 		json_object_set_string_member(properties, "referrer", "https://discord.com/channels/@me");
 		json_object_set_string_member(properties, "referring_domain", "discord.com");
 		json_object_set_string_member(properties, "referrer_current", "");
@@ -1191,7 +1191,7 @@ discord_send_auth(DiscordAccount *da)
 		json_object_set_string_member(properties, "release_channel", "stable");
 		json_object_set_int_member(properties, "client_build_number", 83364);
 		json_object_set_null_member(properties, "client_event_source");
-		
+
 		json_object_set_object_member(data, "properties", properties);
 
 		/* TODO real presence */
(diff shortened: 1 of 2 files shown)


libdiscord.c
```diff
@@ -1737,6 +1737,54 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
...
+	time_t timestamp;
+} DiscordImgMsgContext;
+
+
+static void
+discord_download_image_cb(DiscordAccount *da, JsonNode *node, gpointer user_data) {
+	//The returned size can be changed by appending a querystring of ?size=desired_size to the URL. Image size can be any power of two between 16 and 4096.
+	DiscordImgMsgContext *img_context = user_data;
+	int img_id = -1;
+  gsize img_data_len;
```
EionRobb, line 1754: Space instead of tab?

libdiscord.c
```diff
@@ -1737,6 +1737,54 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
...
+		JsonObject *response = json_node_get_object(node);
+		img_data_s = g_dataset_get_data(node, "raw_body");
+		img_data_len = json_object_get_int_member(response, "len");
+		img_data = g_memdup(img_data_s, img_data_len);
+  	img_id = purple_imgstore_add_with_id(img_data, img_data_len, NULL);
+
+    if (img_id >= 0) {
+      attachment_show = g_strdup_printf("<img id=\"%u\" alt=\"%s\"/><br /><a href=\"%s\">(link)</a>", img_id, img_context->url, img_context->url);
+		} else {
+      attachment_show = g_strdup(img_context->url);
```
EionRobb, line 1769: `attachment_show` needs to be `g_free()`d

libdiscord.c
```diff
@@ -1737,6 +1737,54 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
 	return g_strdup(who);
 }
 
+typedef struct _DiscordImgMsgContext {
+	gint conv_id;
+	gchar* from;
+	const gchar* url;
```
EionRobb, line 1743: This should be made non-const since you do a `g_strdup` when populating it

libdiscord.c
```diff
@@ -1737,6 +1737,54 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
...
+    	purple_serv_got_im(da->pc, img_context->from, attachment_show, img_context->flags, img_context->timestamp);
+		}
+		g_free(img_context);
+		return;
+	} else {
+		purple_debug_error("discord", "Image response node is null!\n");
+		g_free(img_context);
+		return;
+	}
+	g_free(img_context);
```
EionRobb, line 1784: You'll need to `free` all the other parts of the `img_context` that you've copied to get in there, such as the `from` and the `url`.  You could make a helper function for freeing it?

libdiscord.c
```diff
@@ -2058,9 +2106,43 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
...
+							}
+
+							conv = PURPLE_CONVERSATION(imconv);
+						}
+
+						discord_fetch_url(da, img_context->url, NULL, discord_download_image_cb, img_context);
+						GList *l = conv->logs;
+						if (l != NULL) {
+							PurpleLog *log = l->data;
+							printf("%s\n", log->name);
```
EionRobb, line 2138: Having `printf`s in the code will cause display errors on terminal clients such as Finch

libdiscord.c
```diff
@@ -2092,10 +2174,41 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
...
+        	PurpleChatConversation *chatconv = purple_conversations_find_chat(da->pc, discord_chat_hash(channel_id));
+       		conv = PURPLE_CONVERSATION(chatconv);
+
+					int head_count = guild ? g_hash_table_size(guild->members) : 0;
+					if (head_count > 0 && (head_count < purple_account_get_int(da->account, "large-channel-count", 20) || purple_account_get_bool(da->account, "display-images-large-servers", FALSE) )) {
+						discord_fetch_url(da, img_context->url, NULL, discord_download_image_cb, img_context);
+						GList *l = conv->logs;
+						if (l != NULL) {
+							PurpleLog *log = l->data;
+							printf("%s\n", log->name);
```
EionRobb, line 2201: This one removed too, please :)

libdiscord.c
```diff
@@ -2092,10 +2174,41 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
...
+				if (g_str_has_prefix(type, "image") && (!strstr(url, "/SPOILER_")) && purple_account_get_bool(da->account, "display-images", FALSE)) {
+
+					DiscordImgMsgContext *img_context = g_new0(DiscordImgMsgContext, 1);
+					img_context->conv_id = discord_chat_hash(channel_id);
+					img_context->from = g_strdup(name);
+					img_context->url = g_strdup(url);
+					img_context->flags = flags | PURPLE_MESSAGE_IMAGES;
+					img_context->timestamp = timestamp;
+
+        	PurpleChatConversation *chatconv = purple_conversations_find_chat(da->pc, discord_chat_hash(channel_id));
```
EionRobb, line 2192: Some indenting inconsistency's here?  Can you make sure it's tab's everywhere?

EionRobb:
With the other markdown code, we've made sure that there's two way conversion from HTML <-> markdown, so that whatever you type into Pidgin is exactly what's presented to the other person, but if you were to type || into a conversation (expecting that it'll display literally as || to the other person) then it has different behaviour

How about instead, treating black-on-black text as the spoiler flag?

dlmarquis:
> How about instead, treating black-on-black text as the spoiler flag?

I can do that, but given there's no keybinding for the toggle bgcolor function, setting it through the gui is going to feel like a downgrade from the current behavior. Is there a convenient way to set up a keybinding for `gtk_imhtml_toggle_backcolor()`? I'm not too familiar with that part of the codebase.

EionRobb:
Thanks for tweaking the PR :)

alexolog:
I've been using `||` for spoilers for a long time.  It works the same as in the official client.
I really hope that it would continue to work that way.

gatlin:
@dlmarquis thanks for picking up my slack! And @EionRobb thanks for this excellent plugin and for upholding standards of quality.

PR #338, opened 2021-05-12
Display message reactions

dlmarquis:
This adds handling to allow pidgin to display a system message when someone reacts to a message.

The reaction object retrieved from individual messages changed out from under me while coding this, so reaction processing for backlogged messages doesn't currently include info on who is reacting.

Reactions made while the conversation is closed and pidgin is open simply get eaten. The alternative to this would be to open the conversation for a reaction. I'm not sure which behavior would be best here.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -251,6 +251,12 @@ typedef struct _DiscordImgMsgContext {
 	time_t timestamp;
 } DiscordImgMsgContext;
 
+typedef struct {
+	PurpleConversation *conv;
+	gchar *reactor;
+	gchar *reaction;
+} DiscordReaction;
+
 static guint64
 to_int(const gchar *id)
 {
@@ -469,6 +475,15 @@ discord_free_image_context(gpointer data)
 	g_free(img_context->from);
 	g_free(img_context->url);
 	g_free(img_context);
+}
+
+static void
+discord_free_reaction(gpointer data)
+{
+	DiscordReaction *react = data;
+	g_free(react->reactor);
+	g_free(react->reaction);
+	g_free(react);
 }
 
 /* updating */
@@ -1754,6 +1769,61 @@ bail:
 	return g_strdup(who);
 }
 
+static void discord_react_cb(DiscordAccount *da, JsonNode *node, gpointer user_data);
+
+static gchar *
+discord_get_react_text(PurpleConversation *conv, JsonArray *reactions, gchar *name) {
+	guint reactions_len = json_array_get_length(reactions);
+	if (reactions_len > 0) {
+		gchar *reaction_str = NULL;
+		gchar *tmp;
+
+		for (guint n = 0; n < reactions_len; n++) {
+			JsonObject *reaction = json_array_get_object_element(reactions, 0);
+			JsonObject *emoji = json_object_get_object_member(reaction, "emoji");
+			const gchar *emoji_id = json_object_get_string_member(emoji, "id");
+			const gchar *emoji_name = json_object_get_string_member(emoji, "name");
+			guint count = json_object_get_int_member(reaction, "count");
+			gboolean is_me = json_object_get_boolean_member(reaction, "me");
+			gchar *reactors;
+
+			if (count > 1)
+				reactors = g_strdup_printf("%u %s", count, "people");
+			else
+				reactors = is_me ? g_strdup("You") : g_strdup(name);
+
+			if (emoji_name != NULL) {
+				gchar *emoji_str = (emoji_id != NULL) ? g_strdup_printf("&lt;:%s:%s&gt;", emoji_name, emoji_id) : g_strdup(emoji_name);
+
+				if (reaction_str != NULL) {
+					tmp = g_strdup_printf("%s<br />%s reacted with %s", reaction_str, reactors, emoji_str);
+					g_free(reaction_str);
+					reaction_str = tmp;
+				} else {
+					reaction_str = g_strdup_printf("%s reacted with %s", reactors, emoji_str);
+				}
+
+				g_free(emoji_str);
+			}
+			g_free(reactors);
+		}
+
+		/* Replace <:emoji:id> with emojis */
+		if (reaction_str != NULL) {
+			tmp = g_regex_replace_eval(emoji_regex, reaction_str, -1, 0, 0, discord_replace_emoji, conv, NULL);
+
+			if (tmp != NULL) {
+				g_free(reaction_str);
+				reaction_str = tmp;
+			}
+		}
+
+		return reaction_str;
+	}
+
+	return NULL;
+}
+
 static void
 discord_download_image_cb(DiscordAccount *da, JsonNode *node, gpointer user_data) {
 	//The returned size can be changed by appending a querystring of ?size=desired_size to the URL. Image size can be any power of two between 16 and 4096.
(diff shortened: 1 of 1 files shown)


libdiscord.c
```diff
@@ -1754,6 +1769,61 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
...
+			if (count > 1)
+				reactors = g_strdup_printf("%u %s", count, "people");
+			else
+				reactors = is_me ? g_strdup("You") : g_strdup(name);
+
+			if (emoji_name != NULL) {
+				gchar *emoji_str = (emoji_id != NULL) ? g_strdup_printf("&lt;:%s:%s&gt;", emoji_name, emoji_id) : g_strdup(emoji_name);
+
+				if (reaction_str != NULL) {
+					tmp = g_strdup_printf("%s<br />%s reacted with %s", reaction_str, reactors, emoji_str);
```
EionRobb, line 1799: Can you mark these strings as translatable?  To do that, wrap them with the `_()` helper function.  That signals to translators that they can edit them in their own language in the .po files

libdiscord.c
```diff
@@ -1754,6 +1769,61 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
 	return g_strdup(who);
 }
 
+static void discord_react_cb(DiscordAccount *da, JsonNode *node, gpointer user_data);
+
+static gchar *
+discord_get_react_text(PurpleConversation *conv, JsonArray *reactions, gchar *name) {
```
EionRobb, line 1775: Unless you're going to modify `name` or use it as an 'out', specify strings as `const` to signal to the caller that the ownership of the memory isnt' going to be transferred (and also helps pick up a few extra types of double-free's at compile time)

libdiscord.c
```diff
@@ -2156,6 +2227,23 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
...
 		}
+
+		if (reactions != NULL) {
+			PurpleIMConversation *imconv;
+			gchar *username = g_hash_table_lookup(da->one_to_ones, channel_id_s);
+			imconv = purple_conversations_find_im_with_account(username, da->account);
+			conv = PURPLE_CONVERSATION(imconv);
+
+			gchar *reaction_str = discord_get_react_text(conv, reactions, username);
+			g_free(username);
```
EionRobb, line 2238: This will free the memory that's in the one-to-ones hash table.   You could potentially `g_strdup()` it when you call the `_lookup` but the way you're using it in `discord_get_react_text()` doesn't look like you need to dup+free it, maybe just mark it as `const` when you look it up?

libdiscord.c
```diff
@@ -2218,6 +2306,19 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
 
 		}
 
+		if (reactions != NULL) {
+			PurpleChatConversation *chatconv = purple_conversations_find_chat(da->pc, discord_chat_hash(channel_id));
+			conv = PURPLE_CONVERSATION(chatconv);
+			gchar *someone = g_strdup("Someone");
+			gchar *reaction_str = discord_get_react_text(conv, reactions, someone);
+			g_free(someone);
```
EionRobb, line 2314: If you do end up changing `discord_get_react_text` to take in a const string for the name, then it'll save you a few lines of code having to dup/free memory :)

libdiscord.c
```diff
@@ -3287,6 +3388,57 @@ discord_process_dispatch(DiscordAccount *da, const gchar *type, JsonObject *data
...
+		if (channel_id_s && g_hash_table_contains(da->one_to_ones, channel_id_s)) {
+			PurpleIMConversation *imconv;
+
+			gchar *username = g_hash_table_lookup(da->one_to_ones, channel_id_s);
+			imconv = purple_conversations_find_im_with_account(username, da->account);
+
+			conv = PURPLE_CONVERSATION(imconv);
+
+			if (user_id == da->self_user_id)
+				user_nick = "You";
```
EionRobb, line 3414: Can you mark this one as translatable too, with `_("You")`?

libdiscord.c
```diff
@@ -4915,6 +5067,71 @@ discord_chat_leave_by_room_id(PurpleConnection *pc, guint64 room_id)
...
+	JsonObject *user_obj = json_object_get_object_member(data, "author");
+	guint64 user_id = to_int(json_object_get_string_member(user_obj, "id"));
+
+	const gchar *msg_text = json_object_get_string_member(data, "content");
+	size_t txt_len = strlen(msg_text);
+	gchar *prev_text;
+
+	// Truncate long messages
+	if (txt_len > 64) {
+		gchar *tmp = g_strndup(msg_text, 64);
```
EionRobb, line 5091: This could potentially cut multi-byte characters in half.   Maybe have a look at the unicode-handling string functions in glib for a safer alternative? https://developer.gnome.org/glib/stable/glib-Unicode-Manipulation.html#g-utf8-strlen

libdiscord.c
```diff
@@ -4915,6 +5067,71 @@ discord_chat_leave_by_room_id(PurpleConnection *pc, guint64 room_id)
...
+
+		DiscordChannel *channel = discord_get_channel_global(da, channel_id_s);
+		DiscordGuild *guild = discord_get_guild(da, channel->guild_id);
+		user_nick = g_hash_table_lookup_int64(guild->nicknames, user_id);
+
+	}
+
+	gchar *react_text;
+	if (user_id != da->self_user_id)
+		react_text = g_strdup_printf("%s reacted with \"%s\" to %s's message: %s", reactor_nick, emoji_name, user_nick, prev_text);
```
EionRobb, line 5114: Translatable here too, please :)

libdiscord.c
```diff
@@ -4915,6 +5067,71 @@ discord_chat_leave_by_room_id(PurpleConnection *pc, guint64 room_id)
...
+	else
+		react_text = g_strdup_printf("%s reacted with \"%s\" to your message: %s", reactor_nick, emoji_name, prev_text);
+
+	/* Replace <:emoji:id> with emojis */
+	if (react_text != NULL) {
```
EionRobb, line 5119: Can `react_text` ever be `NULL`?

libdiscord.c
```diff
@@ -4915,6 +5067,71 @@ discord_chat_leave_by_room_id(PurpleConnection *pc, guint64 room_id)
 	discord_socket_write_json(ya, data);*/
 }
 
+static void
+discord_react_cb(DiscordAccount *da, JsonNode *node, gpointer user_data)
+{
+	JsonArray *datas = json_node_get_array(node);
+	guint len = json_array_get_length(datas);
+	JsonObject *data = json_array_get_object_element(datas, len-1);
```
EionRobb, line 5075: Is it always guaranteed to be the last message, or could there be a thread race that there's been a message or two sent in between?

dlmarquis: The last message here is actually the oldest message; it should match the logic in the history grabbing. We're searching for 5 messages after index `desired_message - 1`, so the oldest message in the array should be the one we want. I did not test for potential thread races, however.

libdiscord.c
dlmarquis: Ah crap, I thought I'd caught all of those. Thanks for your patience with me; memory management is definitely not my strong suit.

libdiscord.c
dlmarquis: This should be fixed now! Let me know if I need to do any additional text processing.

EionRobb:
Thanks :)

PR #342, opened 2021-05-27
Show referenced post with replies

dlmarquis:
Like it says on the tin, this adds a small preview of the replied-to message before a reply message is posted.

I don't really like the formatting of the reply text as it stands now, but I'm not sure how else to implement it. I'd appreciate some input on that.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -1887,6 +1887,7 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
 	time_t timestamp = purple_str_to_time(timestamp_str, FALSE, NULL, NULL, NULL);
 	const gchar *nonce = json_object_get_string_member(data, "nonce");
 	gchar *escaped_content = purple_markup_escape_text(content, -1);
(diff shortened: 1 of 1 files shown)


libdiscord.c
```diff
@@ -2262,6 +2302,40 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
...
+				prev_text = g_strdup(msg_txt);
+			}
+
+			// Formatting could be better. I went with something similar to Discord's
+			// format to make it familiar to the user
+			gchar *reply_txt = g_strdup_printf("<font size=1>┌──@%s: %s</font>", reply_username, prev_text);
+			g_free(prev_text);
+
+			PurpleChatConversation *chatconv = purple_conversations_find_chat(da->pc, discord_chat_hash(channel_id));
+			conv = PURPLE_CONVERSATION(chatconv);
```
EionRobb, line 2333: This should check if `chatconv` is `NULL`.  If it is, then there's no chat conversation open and the conv_write doesn't need to happen

PR #346, opened 2021-06-09
Allow offline messaging

dlmarquis:
Climbing through the pidgin 2.14 source, I noticed there was a protocol option to enable offline messaging. I enabled it, and it seems to work just fine: my messages show up in the official web client.


EionRobb:
Nice find! Thanks!

PR #350, opened 2021-06-22
Update to use api v9

dlmarquis:
I've been preparing to move to api v9 in anticipation of Discord's new thread feature dropping. I don't know if you're also looking at this @EionRobb, but I thought I'd share my work.

There were surprisingly few changes I had to make, but it's possible I missed something in the [change log](https://discord.com/developers/docs/change-log).


EionRobb:
That was the last piece of the puzzle? Nice!  I've been using v9 for a while but didn't want to push it out to others because no users would ever show up - but the permissions changing between ints and strings makes a whole lot of sense.  Thank you :)

Issue #327, opened 2021-02-23
static analysis: cppcheck: mismatched types, null dereference, scope reduction, shadowed variables, unused code


pabs3:
The cppcheck static analysis tool found various issues in the new code and some of the old code:

```
$ cppcheck -j1 --quiet --enable=all --force .
libdiscord.c:975:15: style: The scope of the variable 'cookie_start' can be reduced. [variableScope]
 const gchar *cookie_start;
              ^
libdiscord.c:3675:14: style: Local variable 'roles' shadows outer variable [shadowVariable]
  JsonArray *roles = json_object_get_array_member(member, "roles");
             ^
libdiscord.c:3643:13: note: Shadowed declaration
 JsonArray *roles = json_object_get_array_member(guild, "roles");
            ^
libdiscord.c:3675:14: note: Shadow variable
  JsonArray *roles = json_object_get_array_member(member, "roles");
             ^
libdiscord.c:5336:0: information: Skipping configuration 'PURPLE_VERSION_CHECK=0;purple_notify_error' since the value of 'purple_notify_error' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly. [ConfigurationNotChecked]
  purple_notify_error(da, _("Bad channel type"), _("Cannot join a voice channel as text"), "", purple_request_cpar_from_connection(da->pc));
^
libdiscord.c:5920:0: information: Skipping configuration 'PURPLE_VERSION_CHECK=0;purple_notify_error' since the value of 'purple_notify_error' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly. [ConfigurationNotChecked]
  purple_notify_error(da, _("Unknown user"), message, "", purple_request_cpar_from_connection(da->pc));
^
purple2compat/http.c:1065:12: warning: %x in format string (no. 1) requires 'unsigned int *' but the argument type is 'signed int *'. [invalidScanfArgType_int]
  if (1 != sscanf(line, "%x", &hc->chunk_length)) {
           ^
purple2compat/http.c:3043:6: warning: Either the condition '!url' is redundant or there is possible null pointer dereference: relative_url. [nullPointerRedundantCheck]
 if (relative_url->host) {
     ^
purple2compat/http.c:1274:8: note: Assuming that condition '!url' is not redundant
   if (!url) {
       ^
purple2compat/http.c:1285:38: note: Calling function 'purple_http_url_relative', 2nd argument 'url' value is 0
   purple_http_url_relative(hc->url, url);
[...]

PR #356, opened 2021-07-29
Fix static analysis complaints

pabs3:
The first commit is a regression since commit fbbe0f9.

The second commit addresses one of the issues from #327.


EionRobb:
Thanks again :)

PR #357, opened 2021-07-29
Clean compiled translations too

pabs3:


EionRobb:
Thanks. Will make life a little easier :)

PR #360, opened 2021-09-01
Explicitly link with zlib on platforms other than Windows/macOS.

pabs3:
Both Windows/macOS were already doing this, the code
uses zlib and Rosa Linux was working around this issue.

See-also: https://abf.io/import/purple-discord/raw/rosa2019.1/purple-discord.spec


EionRobb:
Makes sense.  Thanks :)

PR #365, opened 2021-09-16
Fixed history grabbing on manual join

### LINKED ISSUE #319: chat history?
According to #194 this plugin downloads the chat history on startup. Do I need to manually enable this feature somewhere as no history at all is showing for me.

### LINKED ISSUE #335: discord not pulling missed offline history
I have pidgin 2.14.3 set as a daemon and autorespawning due to this [bug](https://issues.imfreedom.org/issue/PIDGIN-17441) so it is not frequent that this occurs to me. It is always reproducible if I say something in the cell version of Discord, and then bring pidgin online afterwards. No missed messages appear. I can state that when using Bionic Beaver, missed telegram and discord messages would appear as expected. I am using ubuntu 20.04.2 right now. 

Building from source vs package install made no difference. 

I do not know, if  any, link exists between telegram and discord missed messages not being pulled.

What would be useful for me to contribute? I assume I shall start with a backtrace?

dlmarquis:
This is the pr for what we discussed on discord.

I also added a channel setting for how much history to grab (limited to 100 messages or all), because channel traffic isn't well-predicted by guild size.

Now that I think of it, it might make sense to make that setting a trinary switch, with an additional "don't grab history at all" option.

Fixes #335, fixes #319

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -5770,6 +5770,29 @@ discord_got_history_static(DiscordAccount *da, JsonNode *node, gpointer user_dat
 	}
 }
 
+static gboolean
+discord_get_room_history_limiting(DiscordAccount *da, guint64 id)
+{
+	PurpleBlistNode *blistnode = NULL;
+	gchar *channel_id = from_int(id);
+
+	if (g_hash_table_contains(da->one_to_ones, channel_id)) {
+		/* is a direct message */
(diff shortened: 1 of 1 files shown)


EionRobb:
Grabbing *all* history sounds scary.  I like it!

:D

Thanks again for another great pull request

dlmarquis:
And thank you for being patient as I learn proper memory management!

PR #370, opened 2021-09-26
Clean up some compiler warnings

pabs3:


EionRobb:
I'd be worried about us sending >2GB image images files, but sure, can't hurt to be careful 😄

PR #372, opened 2021-09-29
Reply/react support proof-of-concept

dlmarquis:
This is a _very_ bare-bones implementation of outgoing replies and reactions. I don't expect this to get merged, but I wanted to make my work available to others.

This adds a user setting to append a small period containing a link of format `dscmsg:[message id]`, which can then be copied for use in a slash command such as `/reply [message id] [your message content]` or `/react [message id] [emoji]`.

I didn't want to add the gui dependencies to purple-discord proper, so I initiated the `dscmsg` protocol in a separate repo at [dlmarquis/pidgin-discordhelper](https://github.com/dlmarquis/pidgin-discordhelper). This makes the small links clickable, and clicking them copies the message id to the clipboard.

Replying and reacting are currently only enabled for chats, because implementing them for dms was going to be a lot more work.

I've been using this for a while, and it works very well. It occasionally shows the wrong message preview for a replied-to message when I reply to a message from my log history and not the current session, but that's about it for bugs I've noticed.

EDIT 04OCT2021: I added timestamp support that works with the tiny links added to messages. I'm fairly sure it picks up the _last_ message at a given timestamp, if you manage to have multiple. No support for changing that yet.

As long as we're confident our timestamp code works now, I can also have it just pull messages from the server based on the timestamp -> message id conversion.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -74,8 +74,6 @@
 #define DISCORD_CDN_SERVER "cdn.discordapp.com"
 
 #define DISCORD_EPOCH_MS	1420070400000
-#define DISCORD_EPOCH_S	1420070400
-#define DSCMSGSIZE  (sizeof("dscmsg:") - 1)
 
 #define DISCORD_MESSAGE_NORMAL (0)
 #define DISCORD_MESSAGE_EDITED (1)
@@ -331,6 +329,7 @@ typedef struct _DiscordImgMsgContext {
 	gchar* url;
 	PurpleMessageFlags flags;
 	time_t timestamp;
+	guint64 msg_id;
 } DiscordImgMsgContext;
 
 typedef struct {
@@ -338,6 +337,17 @@ typedef struct {
 	gchar *reactor;
 	gchar *reaction;
 } DiscordReaction;
+
+typedef struct {
+	guint64 room_id;
+	gchar *msg_txt;
+	PurpleConversation *conv;
+} DiscordReply;
+
+typedef struct {
+	guint64 msg_id;
+	time_t when;
+} DiscordMsgTimestamp;
 
 static guint64
 to_int(const gchar *id)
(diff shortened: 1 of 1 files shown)


EionRobb:
That sounds interesting!  I'd quite like to merge it.

Do you have screenshots/video of it in action?

dlmarquis:
Here's a gif of me reacting to a message. It's not super interesting, but I hope it's what you want. We could get fancier with the gui by making the link a button, or by having the link open up a separate dialog for replying/reacting when clicked instead of just copying the message id to the clipboard. I'm still not super confident with the gui end of things, so I didn't bother for my own use.

I copy/paste a unicode emoji in this gif, but custom emoji also work by using the :emoji: shortcut, as in regular messages.

reactrecord

This pr really does need the helper plugin, because otherwise pidgin refuses to make the link clickable at all, and you can't recover the message id. If we merge this in, we'll probably want to merge that in here somehow.

EionRobb:
Just trying to think of ideas to get around the helper, and make it work for other frontends:
* Could we record the message id of the most recent message in the convo and use `/react` without the id?
* Could we also react by username instead of message id?

dlmarquis:
I don't think it makes sense to bind it to the most recent message. Often, you want to use the reply feature to reference a message several messages back. By username is less restrictive, but I think we hit a race case if that user posts a new message while we're replying.

I considered tying it to the PurpleConversation's message_history list in some way, making a map between the message ids and the messages in the history, but figured out the current way first.

Now that I think about it, the helper plugin (or equivalent) may only be necessary on pidgin. Pidgin deactivates links it thinks are invalid, but that might not be true for other frontends. My very first working iteration of this code had the small links be in the format `ftp://[message id]`, and I'd just select "copy link location" from the right-click menu and delete the "ftp://" after pasting it into the entry box.

dlmarquis:
Okay, I've given it some more thought, and I think if we store a list of message ids that match up to the PurpleConversation's message_history GList, like I was originally thinking, we could then have users specify messages using the timestamp. They would have to have timestamps turned on for it to work, but I think it's a better compromise than limiting what messages users can reply and react to.

So if the message history looks like this (small links expanded for clarity):
```
(15:17:04) UncreativeName: /beɪɡəl/ gang .(dscmsg:893214887264792638)
(15:22:59) Aspie: /beɡl̩/ :) .(dscmsg:893216376561500250)
(15:23:50) Aspie: Perso je trouve le /ɪ/ est difficile à prononcer parfaitement .(dscmsg:893216589162373201)
(15:25:15) UncreativeName: i could probably do away with the schwa in my transcription lol .(dscmsg:893216949683748974)
```

The user could type

```
/react 15:22:59 🥖
```

and that would be equivalent to

```
/react 893216376561500250 🥖
```

If there are multiple messages with the same timestamp, we could add an extra argument such that `/react 15:22:59 1 🥖` reacts to the first message, `/react 15:22:59 2 🥖` to the second, etc.

I think that would actually still work even keeping the small link in each message and not storing a comparator list, if we wanted to do it that way.

dlmarquis:
Following up on this, I've implemented my idea for storing a list of ids and timestamps with each conversation. The only thing I don't have yet is the extra argument for the case where there are multiple messages sent within one second. I haven't seen that scenario happen yet in my testing though, so hopefully it's a non-issue.

PR #378, opened 2021-11-12
Rudimentary Ratelimiting, QoL Tweaks

dlmarquis:
From #376:

> The biggest change is some rudimentary ratelimiting that allows us to finally query missed group chat messages at startup. Experimentally, Discord has a 5-per-5-second limit on any given endpoint group (per channel, per guild, per webhook). The only place we run into that enough to cause serious problems is history fetching, so I added a 1-second delay before it fetches the next batch of messages. I also added a 30ms delay before any http fetch towards the end of not hitting the global ratelimit. The fetch process should also now abort and resend the request if it receives a code 429 ratelimit exceeded in the response header.
>
>To make the history fetching experience a little better, I added channel settings that allow you to override whether a single channel is treated as large or small.
>
>I also updated some enums, and added message type and permission flag enums to make the code a little less obtuse.
>
>Outside of that, there are a few stray fixes to errors I've made in previous PRs.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -100,8 +100,40 @@ typedef enum {
 	CHANNEL_DM = 1,
 	CHANNEL_VOICE = 2,
 	CHANNEL_GROUP_DM = 3,
-	CHANNEL_GUILD_CATEGORY = 4
+	CHANNEL_GUILD_CATEGORY = 4,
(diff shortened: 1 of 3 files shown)


EionRobb:
Thanks for making the changes to the PR :)

PR #380, opened 2021-11-15
Rename `JsonArray *data` to something else

adrianmalacoda:
This is so as to not conflict with `JsonObject *data`, which results in the following error:

```
libdiscord.c: In function ‘discord_react_cb’:
libdiscord.c:5491:14: error: conflicting types for ‘data’
  JsonObject *data = json_array_get_object_element(data, len-1);
              ^~~~
libdiscord.c:5489:13: note: previous definition of ‘data’ was here
  JsonArray *data = json_node_get_array(node);
             ^~~~
```


EionRobb:
Good catch.  Thanks!

PR #382, opened 2021-11-23
More tweaks

dlmarquis:
- Added an Opcode enum so communication with the gateway is more readable
- Added a RelationshipType enum so that code is more readable as well
- Moved message truncation into its own function to reduce duplicated code
- Added detection for when a user is on mobile
- Added option to fetch a smaller version of an image, because some of my servers love filling my harddrive with absolutely massive cat pictures
- Changed capabilities to 125 from 61 to match the official client
- Added new lines to a few print statements, made at least one string translatable, and corrected a few places that used ints instead of enum names

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -72,6 +72,8 @@
 #define DISCORD_API_VERSION "v9"
 #define DISCORD_CDN_SERVER "cdn.discordapp.com"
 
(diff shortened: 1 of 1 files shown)


EionRobb:
These changes are superlative!  Thank you for tidying up :)

PR #389, opened 2022-01-10
Threads, reacts, replies, and more

dlmarquis:
This is a behemoth of a PR, and I apologize for that. Adding support for threads, reactions, and replies all relied on the same core code, so I figured I'd package it all together, and then a few smaller items I was looking at also got absorbed into this PR due to shared code. Let me know if you want me to split things up or reorganize the commits.

Much of the thread code was ripped from and/or inspired by the thread support over at [slack-libpurple](https://github.com/dylex/slack-libpurple).

Threads are shown in their parent channel's chat window, but I have stored them as full channels so that we can support breaking them out into their own entities if we want to in the future.

Attempting to reply or react within a thread should not currently work.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -20,6 +20,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <time.h>
 #ifdef __GNUC__
 #include <unistd.h>
 #endif
@@ -5640,6 +5641,59 @@ discord_react_cb(DiscordAccount *da, JsonNode *node, gpointer user_data)
 	discord_free_reaction(react);
 }
 
+gint discord_msg_time_comp(gpointer pmsg, gpointer ptime)
+{
+	time_t given_time = *(time_t*)ptime;
+	PurpleConvMessage *msg = (PurpleConvMessage*)pmsg;
+	time_t msg_time = msg->when;
+
+	return !(msg_time == given_time);
+}
+
+static time_t
+discord_parse_timestamp(const gchar *timestring)
+{
+	time_t timestamp;
+	gchar *verified_timestring;
+	GTimeZone *local_time = g_time_zone_new_local();
+	GDateTime *now = g_date_time_new_now_local();
+	gint year = 1970, month = 1, day = 1;
+
+	if (!strchr(timestring, ' ') && !strchr(timestring, 't') && !strchr(timestring, 'T')) {
+		// no separator, so no date attached
+		g_date_time_get_ymd(now, &year, &month, &day);
+
+		verified_timestring = g_strdup_printf("%i-%02i-%02iT%s", year, month, day, timestring);
+	} else {
+		verified_timestring = g_strdup(timestring);
+	}
+
+	GDateTime *msg_time = g_date_time_new_from_iso8601(verified_timestring, local_time);
+	printf("%s \n", g_date_time_format_iso8601(msg_time));
+
+	g_free(verified_timestring);
+
+	if (msg_time == NULL) {
+		return 0;
+	}
+
+	if (g_date_time_difference(msg_time, now) > 0) {
+		// msg_time is in the future, user is probably up past midnight
+		GDateTime *temp = g_date_time_add_days(msg_time, -1);
+		g_date_time_unref(msg_time);
+		msg_time = temp;
+
+		if (g_date_time_difference(msg_time, now) > 0)
+			return 0;
+	}
+
+	timestamp = g_date_time_to_unix(msg_time);
+
+	g_time_zone_unref(local_time);
+	g_date_time_unref(msg_time);
+	return timestamp;
+}
+
 static void
 discord_got_pinned(DiscordAccount *da, JsonNode *node, gpointer user_data)
 {
@@ -7731,26 +7785,41 @@ discord_reply_cb(DiscordAccount *da, JsonNode *node, gpointer user_data)
 
 }
 
-static PurpleCmdRet
-discord_cmd_reply(PurpleConversation *conv, const gchar *cmd, gchar **args, gchar **error, gpointer data)
+static gchar *
+discord_get_message_id_from_timestamp(PurpleConversation *conv, const gchar *timestamp)
 {
-	PurpleConnection *pc = purple_conversation_get_connection(conv);
-	DiscordAccount *da = purple_connection_get_protocol_data(pc);
-	guint64 room_id = *(guint64 *) purple_conversation_get_data(conv, "id");
-	gint ret;
+  time_t msg_time = discord_parse_timestamp(timestamp);
 
-	if (pc == NULL || (int)room_id == -1) {
-		return PURPLE_CMD_RET_FAILED;
+	if (!msg_time) {
+		return NULL;
 	}
 
-	//gchar **parts = g_strsplit(args[0], " ", 2);
-	gchar *msg_txt = g_strdup(args[1]);
+  GList *msg_list = conv->message_history;
+  GList *msg_elem = g_list_find_custom(msg_list, &msg_time, (GCompareFunc)discord_msg_time_comp);
+  PurpleConvMessage *msg = msg_elem->data;
 
+  if (msg == NULL) {
+    return NULL;
+  }
+
+	const gchar *text = g_strrstr(msg->what, "dscmsg:");
+	g_return_val_if_fail(text && strlen(text) > DSCMSGSIZE, FALSE);
+	return g_strndup((char*)text + DSCMSGSIZE, 18);
+}
+
+static gboolean
+discord_chat_reply(DiscordAccount *da, PurpleConversation *conv, guint64 room_id, gchar **args)
+{
+
+	gchar *msg_txt = g_strdup(args[1]);
+	gchar *msg_id;
+	gint ret;
 
 	DiscordGuild *guild = NULL;
-	DiscordChannel *channel = discord_get_channel_global_int_guild(da, room_id, &guild);
-	guint64 last_message = channel->last_message_id;
+	//DiscordChannel *channel = discord_get_channel_global_int_guild(da, room_id, &guild);
+	discord_get_channel_global_int_guild(da, room_id, &guild);
 
+	/* Format outgoing message */
 	msg_txt = discord_make_mentions(da, guild, msg_txt);
 
 	if(guild) {
@@ -7762,9 +7831,21 @@ discord_cmd_reply(PurpleConversation *conv, const gchar *cmd, gchar **args, gcha
 		}
 	}
 
-	g_return_val_if_fail(discord_get_channel_global_int(da, room_id), -1); /* TODO rejoin room? */
+	g_return_val_if_fail(discord_get_channel_global_int(da, room_id), FALSE); /* TODO rejoin room? */
 
-	ret = discord_conversation_send_message(da, room_id, msg_txt, args[0]);
+	/* Get referenced message id */
+	if (strchr(args[0], ':')) {
+		msg_id = discord_get_message_id_from_timestamp(conv, args[0]);
+		if (msg_id == 0) {
+			g_free(msg_txt);
+			return FALSE;
+		}
+	} else {
+		msg_id = g_strdup(args[0]);
+	}
+
+	/* Send message */
+	ret = discord_conversation_send_message(da, room_id, msg_txt, msg_id);
 	if (ret > 0) {
 
 		DiscordReply *reply = g_new0(DiscordReply, 1);
@@ -7772,29 +7853,23 @@ discord_cmd_reply(PurpleConversation *conv, const gchar *cmd, gchar **args, gcha
 		reply->msg_txt = g_strdup(msg_txt);
 		reply->conv = conv;
 
-		gchar *url = g_strdup_printf("https://" DISCORD_API_SERVER "/api/" DISCORD_API_VERSION "/channels/%" G_GUINT64_FORMAT "/messages?limit=5&after=%" G_GUINT64_FORMAT, room_id, last_message-1);
+		gchar *url = g_strdup_printf("https://" DISCORD_API_SERVER "/api/" DISCORD_API_VERSION "/channels/%" G_GUINT64_FORMAT "/messages?limit=5&after=%" G_GUINT64_FORMAT, room_id, to_int(msg_id)-1);
 		discord_fetch_url(da, url, NULL, discord_reply_cb, reply);
 		g_free(url);
 	}
 
 	g_free(msg_txt);
+	g_free(msg_id);
 
-	return PURPLE_CMD_RET_OK;
+	return TRUE;
 }
 
-static PurpleCmdRet
-discord_cmd_react(PurpleConversation *conv, const gchar *cmd, gchar **args, gchar **error, gpointer data)
+static gboolean
+discord_chat_react(DiscordAccount *da, PurpleConversation *conv, guint64 id, gchar **args)
 {
-	PurpleConnection *pc = purple_conversation_get_connection(conv);
-	DiscordAccount *da = purple_connection_get_protocol_data(pc);
-	guint64 id = *(guint64 *) purple_conversation_get_data(conv, "id");
-
-	if (pc == NULL || id == -1) {
-		return PURPLE_CMD_RET_FAILED;
-	}
-
 	//gchar **parts = g_strsplit(args[0], " ", -1);
 
+	gchar *msg_id;
 	const gchar *raw_emoji = args[1];
 	gchar *emoji = NULL;
 	if (g_str_has_prefix(raw_emoji, ":") && g_str_has_suffix(raw_emoji, ":")) {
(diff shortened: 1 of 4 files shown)


libdiscord.c
```diff
@@ -383,6 +405,18 @@ from_int(guint64 id)
...
 }
 
+static time_t
+discord_time_from_snowflake(guint64 id)
+{
+	return (time_t)(((id >> 22) + DISCORD_EPOCH_MS)/1000);
+}
+
+static guint64
+discord_snowflake_from_time(time_t timestamp)
```
EionRobb, line 415: On 32-bit Linux, `time_t` is a 32bit `int`, so *1000 would overflow.  You'd need to convert to a `guint64` first

libdiscord.c
```diff
@@ -1986,6 +2081,132 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
...
+		verified_timestring = g_strdup(timestring);
+	}
+
+	GDateTime *msg_time = g_date_time_new_from_iso8601(verified_timestring, local_time);
+
+
+	g_free(verified_timestring);
+
+	if (msg_time == NULL) {
+		return 0;
```
EionRobb, line 2108: mem leak here.  Need to unref mem for `local_time`, `msg_time` and `now`

libdiscord.c
```diff
@@ -1986,6 +2081,132 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
...
+	}
+
+	if (g_date_time_difference(msg_time, now) > 0) {
+		// msg_time is in the future, user is probably up past midnight
+		GDateTime *temp = g_date_time_add_days(msg_time, -1);
+		g_date_time_unref(msg_time);
+		msg_time = temp;
+
+		if (g_date_time_difference(msg_time, now) > 0)
+			return 0;
```
EionRobb, line 2118: memleak here, need to unref `local_time`, `msg_time` and `now`

libdiscord.c
```diff
@@ -1986,6 +2081,132 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
...
+	GDateTime *now = g_date_time_new_now_local();
+	gint then_year = 1970, then_month = 1, then_day = 1;
+	gint now_year = 1970, now_month = 1, now_day = 1;
+	gchar *format;
+
+	g_date_time_get_ymd(date_time, &then_year, &then_month, &then_day);
+	g_date_time_get_ymd(now, &now_year, &now_month, &now_day);
+
+	if (then_year != now_year || then_month != now_month || then_day != now_day) {
+		format = "(%F %T)";
```
EionRobb, line 2142: Does this need to be localized?

libdiscord.c
```diff
@@ -5555,6 +5912,55 @@ discord_chat_leave_by_room_id(PurpleConnection *pc, guint64 room_id)
...
 }
 
+static void
+discord_thread_parent_cb(DiscordAccount *da, JsonNode *node, gpointer user_data)
+{
+	if (node == NULL) {
+		return;
+	}
+	JsonArray *messages = json_node_get_array(node);
+	guint len = json_array_get_length(messages);
```
EionRobb, line 5922: Need to check if `len == 0`?

libdiscord.c
```diff
@@ -5698,6 +6104,72 @@ discord_chat_pinned(PurpleConnection *pc, int id)
...
+	//g_free(url);
+
+	//gchar *threads_list = g_strdup(_(" Last Message Time  | Name"));
+	gchar *threads_list = g_strdup(_("Active Threads:\n<pre>Creation Time       | Last Message Time   | Name"));
+
+	while (g_hash_table_iter_next(&thread_iter, &key, &value)) {
+		DiscordChannel *thread = value;
+		if (thread) {
+			GDateTime* creation_time = g_date_time_new_from_unix_local(discord_time_from_snowflake(thread->id));
+			gchar *creation_time_s = g_date_time_format(creation_time, "%F %T");
```
EionRobb, line 6145: Needs to be localized?

libdiscord.c
```diff
@@ -7646,6 +8139,411 @@ discord_actions(
...
+discord_get_thread_id_from_timestamp(DiscordAccount *da, PurpleConversation *conv, const gchar *timestamp)
+{
+	guint64 *room_id_ptr = purple_conversation_get_data(conv, "id");
+	if (!room_id_ptr) {
+		return NULL;
+	}
+	guint64 room_id = *room_id_ptr;
+	DiscordChannel *channel = discord_get_channel_global_int(da, room_id);
+
+  time_t thread_time = discord_parse_timestring(timestamp);
```
EionRobb, line 8226: Whitespace issue?

libdiscord.c
```diff
@@ -7646,6 +8139,411 @@ discord_actions(
...
+	g_return_val_if_fail(discord_get_channel_global_int(da, room_id), FALSE); /* TODO rejoin room? */
+
+	DiscordReply *reply = g_new0(DiscordReply, 1);
+	reply->room_id = room_id;
+	reply->msg_txt = g_strdup(msg_txt);
+	reply->conv = conv;
+
+	time_t msg_time;
+	if (strchr(args[0], ':')) {
+  	msg_time = discord_parse_timestring(args[0]);
```
EionRobb, line 8389: whitespace issue?

libdiscord.c
```diff
@@ -7646,6 +8139,411 @@ discord_actions(
...
+		emoji = tmp;
+	}
+
+	if (emoji == NULL) {
+		return FALSE;
+	}
+
+	/* Get referenced message id */
+	if (strchr(args[0], ':')) {
+  	time_t msg_time = discord_parse_timestring(args[0]);
```
EionRobb, line 8441: whitespace issue?

libdiscord.c
```diff
@@ -1986,6 +2081,132 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
...
+	if (!strchr(timestring, ' ') && !strchr(timestring, 't') && !strchr(timestring, 'T')) {
+		// no separator, so no date attached
+		g_date_time_get_ymd(now, &year, &month, &day);
+
+		verified_timestring = g_strdup_printf("%i-%02i-%02iT%s", year, month, day, timestring);
+	} else {
+		verified_timestring = g_strdup(timestring);
+	}
+
+	GDateTime *msg_time = g_date_time_new_from_iso8601(verified_timestring, local_time);
```
EionRobb, line 2102: `g_date_time_new_from_iso8601` is too-new for win32

libdiscord.c
```diff
@@ -3366,6 +3678,22 @@ discord_process_dispatch(DiscordAccount *da, const gchar *type, JsonObject *data
 				purple_chat_conversation_set_topic(chatconv, NULL, new_topic);
 			}
 		}
+	} else if (purple_strequal(type, "THREAD_LIST_SYNC")) {
+		guint64 guild_id = to_int(json_object_get_string_member(data, "guild_id"));
+		DiscordGuild *guild = discord_get_guild(da, guild_id);
+		JsonArray *threads = json_object_get_array_member(data, "threads");
+
+		for (int i = 0; i < json_array_get_length(threads); i++) {
```
EionRobb, line 3686: compile error on windows: "error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]"

libdiscord.c
```diff
@@ -7646,6 +8139,411 @@ discord_actions(
...
+}
+
+static PurpleCmdRet
+discord_cmd_react(PurpleConversation *conv, const gchar *cmd, gchar **args, gchar **error, gpointer data)
+{
+	PurpleConnection *pc = purple_conversation_get_connection(conv);
+	DiscordAccount *da = purple_connection_get_protocol_data(pc);
+	guint64 id = *(guint64 *) purple_conversation_get_data(conv, "id");
+
+	if (pc == NULL || id == -1) {
```
EionRobb, line 8534: compile error on windows: "error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]"

(id is unsigned so can never be negative)

libdiscord.c
```diff
@@ -1986,6 +2081,132 @@ discord_get_real_name(PurpleConnection *pc, gint id, const char *who)
...
+
+	gchar *color_string = g_strdup_printf("%06x", color);
+
+	return color_string;
+
+}
+
+static gchar *
+discord_get_formatted_thread_timestamp(time_t ts) {
+	gchar *color = discord_get_thread_color(ts);
```
EionRobb, line 2202: Need to `g_free` color after finishing with it

libdiscord.c
```diff
@@ -2288,6 +2536,34 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
...
+		const gchar *color = "#606060";
+		const gchar *indicator;
+		if (thread->last_message_id == 0 || msg_id == thread->id) {
+			indicator = purple_account_get_string(da->account, "parent-indicator", "◈ ");
+		} else {
+			indicator = purple_account_get_string(da->account, "thread-indicator", "⤷ ");
+		}
+		// Do this manually because msg counts for parent channel, not thread
+		thread->last_message_id = msg_id;
+		gchar *thread_ts = discord_get_formatted_thread_timestamp(ts);
```
EionRobb, line 2551: need to `g_free` thread_ts after using it

libdiscord.c
```diff
@@ -2661,6 +2955,27 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
...
+			if (conv != NULL) {
+				purple_conversation_write(conv, NULL, new_thread_txt, PURPLE_MESSAGE_SYSTEM, timestamp);
+			}
+			g_free(new_thread_txt);
+
+
+			if (msg_type == MESSAGE_THREAD_STARTER_MESSAGE) {
+				// Get array of messages because single-message endpoint is bot-only
+				gchar *url = g_strdup_printf("https://" DISCORD_API_SERVER "/api/" DISCORD_API_VERSION "/channels/%" G_GUINT64_FORMAT "/messages?limit=5&after=%" G_GUINT64_FORMAT, channel_id, ref_id-1);
+				discord_fetch_url(da, url, NULL, discord_thread_parent_cb, from_int(thread->id));
```
EionRobb, line 2975: can `thread` be `NULL`?

libdiscord.c
```diff
@@ -3366,6 +3678,22 @@ discord_process_dispatch(DiscordAccount *da, const gchar *type, JsonObject *data
...
+		DiscordGuild *guild = discord_get_guild(da, guild_id);
+		JsonArray *threads = json_object_get_array_member(data, "threads");
+
+		for (int i = 0; i < json_array_get_length(threads); i++) {
+			JsonObject *thread = json_array_get_object_element(threads, i);
+			guint64 thread_id = to_int(json_object_get_string_member(thread, "id"));
+			guint64 parent_id = to_int(json_object_get_string_member(thread, "parent_id"));
+			DiscordChannel *parent = discord_get_channel_global_int(da, parent_id);
+
+			if (!g_hash_table_contains_int64(parent->threads, thread_id)) {
```
EionRobb, line 3692: Could `parent` be `NULL`?

libdiscord.c
```diff
@@ -5698,6 +6104,72 @@ discord_chat_pinned(PurpleConnection *pc, int id)
...
+	DiscordChannel *channel = discord_get_channel_global_int_guild(da, room_id, &guild);
+
+	//gchar *url;
+
+	GHashTableIter thread_iter;
+	gpointer key, value;
+	/* TODO: get archived thread data */
+	if (filter && purple_strequal(filter, "guild")) {
+			//g_hash_table_iter_init(&thread_iter, guild->threads);
+			g_hash_table_iter_init(&thread_iter, channel->threads);
```
EionRobb, line 6130: need a null check for `channel`?

libdiscord.c
```diff
@@ -5698,6 +6104,72 @@ discord_chat_pinned(PurpleConnection *pc, int id)
...
+		//url = g_strdup_printf("https://" DISCORD_API_SERVER "/api/" DISCORD_API_VERSION "/guilds/%" G_GUINT64_FORMAT "/threads", guild->id);
+	} else {
+			g_hash_table_iter_init(&thread_iter, channel->threads);
```
EionRobb, line 6133: the else is the same statement as in the if, do these need to be de-dup'd?

libdiscord.c
```diff
@@ -7646,6 +8139,411 @@ discord_actions(
...
+
+  time_t thread_time = discord_parse_timestring(timestamp);
+	if (!thread_time) {
+		return NULL;
+	}
+
+	GHashTableIter iter;
+	gpointer key, value;
+
+	g_hash_table_iter_init(&iter, channel->threads);
```
EionRobb, line 8234: need null-check for `channel`?

libdiscord.c
```diff
@@ -7646,6 +8139,411 @@ discord_actions(
...
+	ret = discord_conversation_send_message(da, to_int(thread_id), msg_txt, NULL);
+	if (ret > 0 && guild) {
+		gchar *name = discord_create_nickname_from_id(da, guild, channel, da->self_user_id);
+
+		/* Handle thread formatting */
+		time_t ts = discord_time_from_snowflake(to_int(thread_id));
+		const gchar *color = "#606060";
+		const gchar *indicator = purple_account_get_string(da->account, "thread-indicator", "⤷ ");
+
+		gchar *thread_ts = discord_get_formatted_thread_timestamp(ts);
```
EionRobb, line 8288: need to `g_free` this after using it

libdiscord.c
```diff
@@ -7646,6 +8139,411 @@ discord_actions(
...
+
+static gchar **
+discord_parse_wS_args(gchar **args)
+{
+	gchar *new = purple_markup_strip_html(args[0]);
+	gchar **in_progress = g_strsplit(new, " ", 2);
+	gchar **matcher = g_strsplit(args[0], " ", -1);
+
+	if (g_strv_length(matcher) < 2)
+		return NULL;
```
EionRobb, line 8472: need to free `in_progress` and `matcher`

EionRobb (changes requested):

libdiscord.c
dlmarquis: If you mean does it need to be set to a local time zone, no. Since the GDateTimes have a time zone associated with them (the local one), this is done automatically.

If you mean should the string be localized for different date formats, I haven't seen any indication that vanilla libpurple/pidgin uses formats that aren't YYYY-MM-DD. I was considering adding a couple preferences to set the date format, like in slack-libpurple, but decided to put that off until after the core got merged in. I can go back and do that now, if you'd prefer.

libdiscord.c
dlmarquis: Right. I was originally going to allow the user to search per-guild for threads due to how the api endpoints work, then I was planning on using the space to specify archived or private threads, and then I completely forgot about that. :facepalm: 

I'll de-dup it for now.

libdiscord.c
dlmarquis: It shouldn't be, but I added a check just in case

libdiscord.c
dlmarquis: Shouldn't have to unref msg_time here, since it's NULL

libdiscord.c
EionRobb: Ah, I'm thinking of how in the Pidgin .po files, there's a translatable string for how dates should be displayed...
```
#.
#. * This string determines how some dates are displayed.  The default
#. * string "%x %X" shows the date then the time.  Translators can
#. * change this to "%X %x" if they want the time to be shown first,
#. * followed by the date.
#.
#, c-format
msgid "%x %X"
msgstr "%x %X"
```

Since this function is used for outputting display dates, should we be doing something like that?  (I really did mean this as a question rather than a passive 'lets do this'.  Very curious for opinions to this query)

libdiscord.c
EionRobb: No harm done :)  just checking to see if there wasn't something missed, rather than it's just leaving room for future dev

libdiscord.c
dlmarquis: Right now the timestring used with the slash commands needs to conform to ISO8601 (so I don't go insane trying to account for every sensible format), and that standard (or glib's implementation thereof) doesn't allow for putting the date after the time.

I think letting users format the output as time-date is going to lead to them trying to type it in as time-date, which is invalid.

libdiscord.c
dlmarquis: I don't know if you saw- I ripped this function from glib into its own header, and added the necessary check in glib_compat.h. I don't have a windows machine readily available, so I can't test how well that works

libdiscord.c
```diff
@@ -2146,15 +2376,16 @@ discord_process_message(DiscordAccount *da, JsonObject *data, unsigned special_t
...
+		purple_debug_info("discord", "No author in message processed\n");
 		return msg_id;
 	}
 
 	JsonObject *author_obj = json_object_get_object_member(data, "author");
 	guint64 author_id = to_int(json_object_get_string_member(author_obj, "id"));
 
-	const gchar *channel_id_s = json_object_get_string_member(data, "channel_id");
-	guint64 channel_id = to_int(channel_id_s);
+	guint64 id = to_int(json_object_get_string_member(data, "channel_id"));
```
EionRobb, line 2386: There's an `id` already being used in line 2459 and line 2474, so this one shadows them - might cause some issues?

EionRobb (changes requested):
Looks great, if you could just rename that `id` then I think we've got a winner :)

PR #390, opened 2022-01-29
Use older g_time_val_from_iso8601 instead of g_date_time_new_from_iso8601 copy

pabs3:
g_date_time_new_from_iso8601 is present since glib 2.56 but older versions had
g_time_val_from_iso8601 to do the same job, so use that instead of the copy,
which is out of date already in the version that was copied.

Add small g_date_time_new_from_iso8601 wrapper of g_time_val_from_iso8601,
(copied from sysprof) so that the new function can be used with old glib.

Keep the wrapper code unmodified from the sysprof version and
use a define to workaround the need for renaming the wrapper.

The copied in code was also missing copyright and license information,
document the source location and copyright for the wrapper function.

Code-from: https://gitlab.gnome.org/GNOME/sysprof/-/raw/master/src/libsysprof-ui/sysprof-details-page.c
Fixes: commit 0b78c85b40bbfa976f18ca4a2e8dafd885fa0f1c


EionRobb (changes requested):
I don't think `g_autoptr()` is available on winpidgin

pabs3:
Thanks, modified the wrapper function to avoid g_autoptr.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #394, opened 2022-02-06
Don't store a variable that isn't used before return

pabs3:
Suggested-by: scan-build
libdiscord.c:3688:6: warning: Value stored to 'channel' is never read [deadcode.DeadStores]
                                        channel = discord_add_thread(da, guild, parent, data, guild->id);
                                        ^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


EionRobb:
You're awesome.
Thanks for the tidy-ups :)

PR #399, opened 2022-02-17
Added a per-guild preferences ui, added ui for guild member screening

dlmarquis:
Two things here, both are very PurpleRequest-heavy:

1. Added a way to manage settings per-guild. It's found under `Accounts->[discord account]->Manage Servers...` in the BList menu. Right now you can change the guild name prefix added to groups during BList auto-generation, override whether the guild is considered large or small, or leave the guild entirely. BList changes occur on next restart.
2. Added a ui for the guild member screening prompt and guild join request updates. Right now, Discord only has one field available (agree to guild rules), but I tried to write it to be forward-compatible with their proposed updates, which include multiple fields. Discord stores the guild rules as a series of labels attached to the agree-to-rules checkbox field, which is why I was messing around with labels and wrapping earlier. I expect future updates from Discord to make text-wrapped labels necessary in some way.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -86,6 +86,10 @@
 #define DISCORD_MESSAGE_EDITED (1)
 #define DISCORD_MESSAGE_PINNED (2)
 
+#define DISCORD_GUILD_SIZE_DEFAULT (0)
+#define DISCORD_GUILD_SIZE_LARGE (1)
+#define DISCORD_GUILD_SIZE_SMALL (2)
+
 #define IGNORE_PRINTS
 
 static GRegex *channel_mentions_regex = NULL;
@@ -2539,11 +2543,11 @@ discord_treat_room_as_small(DiscordAccount *da, guint64 room_id, DiscordGuild *g
 		return TRUE;
 	}
 	gchar *sizepref_id = g_strdup_printf("%" G_GUINT64_FORMAT "-size", guild->id);
-	gint sizepref = purple_account_get_int(da->account, sizepref_id, 0);
+	gint sizepref = purple_account_get_int(da->account, sizepref_id, DISCORD_GUILD_SIZE_DEFAULT);
 	g_free(sizepref_id);
-	if (sizepref == 1) {
+	if (sizepref == DISCORD_GUILD_SIZE_LARGE) {
 		return FALSE;
-	} else if (sizepref == 2) {
+	} else if (sizepref == DISCORD_GUILD_SIZE_SMALL) {
 		return TRUE;
 	}
 	if (g_hash_table_size(guild->members) < purple_account_get_int(da->account, "large-channel-count", 20))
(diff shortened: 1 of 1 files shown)


libdiscord.c
```diff
@@ -8421,6 +8502,104 @@ discord_add_account_options(GList *account_options)
...
+	}
+
+	JsonArray *json_fields = json_object_get_array_member(json_form, "form_fields");
+	gint form_len = json_array_get_length(json_fields);
+	for (gint n = 0; n < form_len; n++) {
+		JsonObject *json_field = json_array_get_object_element(json_fields, n);
+		gchar *id = g_strdup_printf("field-%d", n);
+		PurpleRequestField *field = purple_request_fields_get_field(fields, id);
+		PurpleRequestFieldType type = purple_request_field_get_type(field);
+		switch (type) { // This will get longer whenever Discord updates this stuff
```
EionRobb, line 8525: Looks like an unclosed } ?

libdiscord.c
```diff
@@ -4288,6 +4304,37 @@ discord_process_dispatch(DiscordAccount *da, const gchar *type, JsonObject *data
...
+			const gchar *rejection_reason = json_object_get_string_member(request, "rejection_reason");
+			if (rejection_reason == NULL) {
+				// Probably pending
+				info = g_strdup_printf(_("Your request to join the server %s is currently pending. You will be notified of any updates regarding your request."), guild->name);
+			} else {
+				info = g_strdup_printf(_("Your request to join the server %s was rejected. The reason given was:\n\n%s"), guild->name, rejection_reason);
+			}
+		}
+
+		purple_request_action(
```
EionRobb, line 4328: Any reason that the `notify` api couldn't be used for this?  Non-Pidgin UI's support `notify` better than they do `request`

libdiscord.c
EionRobb: Oh, no, it's just the indentation doesn't go in far enough... makes it look like the `default:` has a closing } instead of the `switch`

dlmarquis: For the switch statement?
That last `default:\nbreak;` doesn't have braces, so the } following those lines closes the switch statement. I can add braces to make it clearer.

dlmarquis: I think I borrowed the indentation style from a switch statement already in the file, but I can change that as well if you prefer

libdiscord.c
```diff
@@ -8462,6 +8641,148 @@ discord_join_server(PurpleProtocolAction *action)
...
+
+		/* Handle leaving guild bool */
+		PurpleRequestField *field = prefs->data;
+		gboolean value = purple_request_field_bool_get_value(field);
+
+		if (value == 0) {
+			continue;
+		}
+
+		gchar *guild_id = g_strndup(purple_request_field_get_id(field),18);
```
EionRobb, line 8705: Are guild ids always going to be 18 characters long?

dlmarquis: Thought the answer to this was "yes", but I just re-did my math and it won't be true starting July 22 of this year. I'll make this line a little smarter

PR #400, opened 2022-03-25
Incremental guildmember fetch

dlmarquis:
I've added an option to sanity-check the number of guild members we might fetch, so I think we can merge this in now. I've included a way to disable the sanity check, since I've been able to run pidgin without it without consuming an absurd amount of memory.

The search-based way of fetching guild members still runs, so there's no real hard maximum on tracked guild members. I don't know if that's something we want to run with or not.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -292,6 +292,7 @@ typedef struct {
 	GHashTable *members;	 /* list of member ids */
 	GHashTable *nicknames;	 /* id->nick? */
 	GHashTable *nicknames_rev; /* reverse */
+	guint next_mem_to_sync;
 
 	GHashTable *channels;
 	GHashTable *threads;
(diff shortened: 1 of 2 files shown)


libdiscord.c
```diff
@@ -8586,6 +8624,9 @@ discord_add_account_options(GList *account_options)
 	option = purple_account_option_bool_new(_("Display custom emoji as inline images"), "show-custom-emojis", TRUE);
 	account_options = g_list_append(account_options, option);
 
+	option = purple_account_option_int_new(_("Approximate max number of guild members to keep track of (0 disables)"), "max-guild-presences", 200);
```
EionRobb, line 8627: Naming stuff is hard.  They're called "guilds" and "members" internally to the plugin, but most other people know of them as "servers" and "users".  In a perfect world we wouldn't need to have people looking at the readme to know what the setting does (as they might have gotten the plugin from their package manage which wouldn't have an easy way to get to the readme)
So how about something along the lines of... "Max number of users to watch online status for, per server"?

libdiscord.c
```diff
@@ -4493,6 +4499,16 @@ discord_process_dispatch(DiscordAccount *da, const gchar *type, JsonObject *data
 				}
 			}
 		}
+		guint member_count = json_object_get_int_member(data, "member_count");
+		guint online_count = json_object_get_int_member(data, "online_count");
+		guint max_count = (guint) (purple_account_get_int(da->account, "max-guild-presences", 200) - 1); // turns into MAX_UINT when max-guild-presences=0
```
EionRobb, line 4504: Relying on compiler implementation of overflow is a bit messy.  Could we clean it up with a (if <= 0 then MAX_UINT) kinda thing?

Alcaro: It's not compiler specific, the C standard demands that unsigned integer overflow wraps.

But relying on obscure parts of the C standard is a bit messy, and I agree with changing it.

EionRobb:
Nice :)

PR #416, opened 2022-08-07
Convert single CRLF line ending to LF line ending

pabs3:


EionRobb:
Weird that there's just the one of them.  Does it trigger a compiler error somewhere?

Alcaro:
I've never seen any compiler complain about mixed linebreaks, but some IDEs and other text editors do.

pabs3:
No, I only noticed when doing `git init` in an extract of the tarball,
git said it was converting the line feeds over, so I then did a
`dos2unix` in the regular git repository and it only found one CRLF.

-- 
bye,
pabs

https://bonedaddy.net/pabs3/

PR #426, opened 2022-09-23
fixes build issues

msschulte:
* It's now possible to disable USE_QRCODE_AUTH.
* QRCODE flags are now inserted into CFLAGS & LDFLAGS (WIN32_PIDGIN).


EionRobb:
So did it not disable when building with `USE_QRCODE_AUTH=0 make`?

msschulte:
You are right, it does. I haven't tried it this way.

PR #422, opened 2022-08-30
doc: add link to debian packages

nodiscc:

PR #430, opened 2022-11-13
Replace the Fedora/CentOS/RHEL section with a link to Repology

pabs3:
Repology documents which distros purple-discord is available from.

Keep the Gentoo section because purple-discord isn't available there yet.

Obsoletes: https://github.com/EionRobb/purple-discord/pull/422


EionRobb:
Awesome as always @pabs3🙂 
Thank you again

PR #431, opened 2022-11-15
update to version 2 of remote auth api

iv4nhoe:
Didn't get a QR code with Remote Auth v1 anymore, so I updated to v2. Tested creating a new account, got QR code and logged in successfully.

Resources used:
https://luna.gitlab.io/discord-unofficial-docs/desktop_remote_auth.html (v1)
https://luna.gitlab.io/discord-unofficial-docs/desktop_remote_auth_v2.html


EionRobb:
Sorry for the delay. I thought I had posted a review comment but doesn't look like it went through 

(it just asked if there should have been a connection error instead of just a debug log error)

PR #440, opened 2023-02-15
fix some memory leaks

iv4nhoe:


EionRobb:
Nice find!  Thanks!

Out of interest, were you using a tool to find the leaks?

iv4nhoe:
> Nice find! Thanks!
> 
> Out of interest, were you using a tool to find the leaks?

Yes, I ran in valgrind while trying to track down an invalid read I'm seeing occasionally.

Issue #164, opened 2017-12-27
Support file attachments (uploads)


alexolog:
As the title says.  The current plugin does not support attaching files.

bobbybee:
Any reason you would actually *want* this, as opposed to uploading it
somewhere else and linking (possibly better for privacy as well,
depending how DiscordBot behaves)?

DuendeInexistente:
One can't really denny that just choosing the file and uploading it in-chat
is more practical, for people who doesn't have some shell script to upload
from the terminal or file browser.

bobbybee:
Sure, I can: a Pidgin plugin could intercept files and instead upload to
a selected host, substituting in a link at will.

---

Hey, Ferb, I knew what we're going to do today!

alexolog:
It's a question of convenience.

I can drag a file into the Discord client (or select it as an attachment) and it will upload it to the Discord servers and post a link.  It works pretty well for channels, if the file size is below the limit, and for private conversations, if the files are not treated as actually private.

I guess the most flexible solution would be to support it with an option to specify a different upload host to upload to, as @bobbybee mentioned above.

And yes, it can be done manually, but it is less convenient.

bobbybee:
> I guess the most flexible solution would be to support it with an
> option to specify a different upload host to upload to, as @bobbybee
> mentioned above.

More specifically, that this should be handled in Pidgin (or whatever
people use) rather than in this prpl. I'm not even sure the prpl *could*
implement this in a convenient generic way...

alexolog:
What about a non-generic way?

bobbybee:
I haven't played with Pidgin/libpurple stuff in a little while; you'd
have to ask Eion about that.

DuendeInexistente:
Maybe you could execute a script and redirect its output to the message
box, having a default host be a simple one like catbox.moe which supports
curl upload.

DuendeInexistente:
Oh and, maybe this'd be better as a standalone plugin, I don't see why have
it only work on Discord.

bobbybee:
> Oh and, maybe this'd be better as a standalone plugin, I don't see why have
> it only work on Discord.

^^ Exactly. I'm not opposed at all to the idea -- I'd just like it to
work on XMPP too. No changes required on our end! :)

Michal-Szczepaniak:
@EionRobb thats good :) I would help if i could

TomDataworks:
[...]

PR #452, opened 2023-09-20
Added file upload action

B1G-FUNGUS:
Implimentation of https://github.com/EionRobb/purple-discord/issues/164

Allows users to upload a file of any type to a channel through an action in the "More" menu. The code is pretty similar to that of `discord_conversation_send_image`, we just have to probe for the file's info/contents ourselves because we can't use the `purple_img*` commands. Also the nonce for the file upload is not added so we can see the message on our client (there's probably a way to post a fake message on our client so the user gets instant feedback, but this way the user only sees the file upload if it actually succeeds).

Of course, please let me know if there are any issues with the code or style. I've tested this on my machine and I'm relatively confident there should be no issues with the code if it was merged as-is, but the more eyes the merrier.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -9730,10 +9730,7 @@ discord_cmd_get_history(PurpleConversation *conv, const gchar *cmd, gchar **args
 	return PURPLE_CMD_RET_OK;
 }
 
-// TODO positioning
-
-// This function may seem pointless now but if we ever decide to add proper
-// progress indication in the future it may be useful
+// This function may seem pointless now but it might be useful 
 static void
 discord_xfer_finish(PurpleXfer *xfer) {
 	g_free(xfer->data);
@@ -9742,7 +9739,6 @@ discord_xfer_finish(PurpleXfer *xfer) {
 
 // TODO we could do a bunch of other stuff like setting the file size &
 // thumbnail, though that is of questionable utility for an http upload
-
 static void
 discord_xfer_send_init(PurpleXfer *xfer)
 {
@@ -9812,12 +9808,15 @@ discord_xfer_send_init(PurpleXfer *xfer)
 	// help other programs guess at what we're doing
 	purple_xfer_start(xfer, 0, url, 443);
 
-	// TODO we could edit/fork this function to give us progress/completion
-	// notifications, etc
+	// TODO we could edit this function or add our own code for stuff like
+	// progress, completion, the ability to actually cancel the transfer,
+	// etc.
 	discord_fetch_url_with_method_len(da, "POST", url, postdata->str, postdata->len, NULL, NULL);
 
-	purple_xfer_unref(xfer);
+	// TODO We could also wait until we receive the url on our client before
+	// ending our xfer
 	purple_xfer_set_completed(xfer, TRUE);
+	purple_xfer_unref(xfer);
 
 	g_free(filename);
 	g_free(url);
@@ -9845,9 +9844,8 @@ discord_create_xfer(PurpleConnection *pc, guint64 room_id, const gchar *receiver
 	xfer->data = room_id_ptr;
 
 	purple_xfer_set_init_fnc(xfer, discord_xfer_send_init);
-	// TODO check if we need the other functions
 	purple_xfer_set_end_fnc(xfer, discord_xfer_finish);
-	// TODO manually cancelling will NOT stop the xfer
+	// manually canceling the transfer will not actually stop it
 	purple_xfer_set_cancel_send_fnc(xfer, discord_xfer_finish);
 
 	return xfer;
@@ -9869,16 +9867,11 @@ discord_send_file(PurpleConnection *pc, const gchar *who, const gchar *filename)
 
 	PurpleXfer *xfer = discord_create_xfer(pc, room_id, who);
 
-	// TODO do we really need to check for a null byte the way we are
-	// implementing this? I'd assume so, but the jabber prpl doesn't do
-	// this so I better double check
 	if (filename && *filename)
 		purple_xfer_request_accepted(xfer, filename);
 	else
 		purple_xfer_request(xfer);
 }
-
-// TODO should we use the discord hash function for any of this?
 
 static void
 discord_chat_send_file(PurpleConnection *pc, int id, const gchar *filename) {
@@ -9913,8 +9906,9 @@ discord_can_receive_file(PurpleConnection *pc, const gchar *who)
 
 static gboolean
 discord_chat_can_receive_file(PurpleConnection *pc, int id) {
-	// TODO i don't believe there are any checks we can perform here that
-	// aren't better performed later on
+	// As of now, I don't forsee any conditions under which we'd like to
+	// prematurely disable the save function, but I'll leave this here just
+	// in case
 	return TRUE;
 }
 
@@ -10183,8 +10177,6 @@ plugin_init(PurplePlugin *plugin)
 	prpl_info->chat_send_file = discord_chat_send_file;
 	prpl_info->chat_can_receive_file = discord_chat_can_receive_file;
 	prpl_info->can_receive_file = discord_can_receive_file;
-	// TODO ????
-	// prpl_info->new_xfer = discord_new_xfer;
 
 	prpl_info->roomlist_get_list = discord_roomlist_get_list;
 	prpl_info->roomlist_room_serialize = discord_roomlist_serialize;
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -434,6 +434,11 @@ typedef struct {
 	PurpleConversation *conv;
 } DiscordReply;
 
+typedef struct {
+	guint64 room_id;
+	gboolean xfer_started;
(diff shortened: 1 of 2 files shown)


B1G-FUNGUS:
I'm also working on adding this option for direct messages but I'm not sure if that'd be better as it's own merge request or as a part of this one.

B1G-FUNGUS:
IM compatabiltiy has now been added. I also changed the menu item to a command. It may seem a little silly at first, but it seems to fit better and simplifies the code by avoiding issues around getting the conversation and it's id. I could compartmentalize the code a little more since a lot of it is shared, but Pidgin 3.0 is planning on adding file attachments of any type anyways, so I'm trying to avoid doing anything which is going to be undone by that.

Anyways, I'll be testing the command for a bit more to make sure I haven't missed anything, and I'll un-draft the PR then.

alexolog:
Pidgin 3.0 will take a long while to get here

libdiscord.c
```diff
@@ -9730,6 +9730,112 @@ discord_cmd_get_history(PurpleConversation *conv, const gchar *cmd, gchar **args
...
+{
+	PurpleAccount *acct = purple_conversation_get_account(conv);
+	PurpleConnection *pc = purple_account_get_connection(acct);
+	DiscordAccount *da = purple_connection_get_protocol_data(pc);
+
+	gchar *who = NULL;
+	if  (conv->type == PURPLE_CONV_TYPE_IM) who = conv->name;
+
+	// TODO is da the right handle for this or should we use pc?
+	purple_request_file(da, _("Select File To Send"), NULL, FALSE, PURPLE_CALLBACK(discord_upload_ok_cb), NULL, acct, who, conv, conv);
```
EionRobb, line 9834: As your TODO suggests yes it should be the `pc` :)

When the account is closed from the UI, libpurple will automatically stop anything associated with that `handle` parameter (for other non-protocol, you'd specify the PurplePlugin) so that there isn't any activity happening on a closed account/connection (or disabled plugin)

EionRobb:
I like where this PR is going :)

If you want to take it to the next level (and feel free to pass on this, we can always add it in in a separate PR) you can hook into the [`prpl_info->send_file()`](https://docs.imfreedom.org/pidgin2/struct__PurplePluginProtocolInfo.html#aab5dbd311165214be56244f6e6e3c521) and [`prpl_info->send_chat_file()`](https://docs.imfreedom.org/pidgin2/struct__PurplePluginProtocolInfo.html#aeb0d946b327a8553228e1ea6ffe22c83) which will let users drag and drop files onto the conversation window (if they're using Pidgin), and also opens up compatibility with other non-Pidgin libpurple clients that send files

An [example of implementing this API in the Teams plugin](https://github.com/EionRobb/purple-teams/blob/9065c077808497afdc3d813b83a7f65de9f16ae8/teams_contacts.c#L813-L915) if that's of any help

alexolog:
Please do!

B1G-FUNGUS:
Thanks for the info & example! I'm going to go ahead with adding the send file hooks

B1G-FUNGUS:
Alright, status update! Firstly, I've adapted the upload functions for the  xfer hooks. Again I'll hold onto it for a bit to make sure I've ironed everything out. Anyways, here are some notes/concerns:

#### Libpurple 3 Compatibility

I've decided not to write this addition with version 3 compatibility in mind because a) I just don't have time to test that stuff and b) if version 3 adds generic file attachments a lot of the code should be rewritten anyways to use the attachment interface rather than the xfer interface (which is designed for p2p transfers). That said, if someone else wants to they can implement compatibility features.

#### Technical Issues

The only issue I have right now is that pidgin doesn't close the transfer window when the last xfer has been canceled, and I'm not sure if that is a pidgin issue or if I'm doing something wrong. 

#### Design Issues

There are a lot of features we could implement in an xfer/upload that I'm not sure are worth the time/complexity cost of adding. Thumbnails and new DM creation with an upload seem like a lot of hassle for little utility.  Being able to accurately see upload progress and cancel an upload midway is more useful (a 25MB upload can take a bit on a slow connection). That said, the former is mostly visual, the latter doesn't even consistently work on the official Discord client, and both would likely require serious changes to `discord_fetch_url_with_method_len` or a creation of a new function entirely.

Anyways, I'd still be willing to implement these features if people are interested in them, I just don't want to add stuff that isn't going to be used.

alexolog:
As I am using the plugin in preference to the native Discord whenever I can, I will definitely be using whatever features you add, and thanking you from the bottom of my heart!

Just please make sure it works on Windows as well.

B1G-FUNGUS:
@alexolog thanks for the reminder. I made sure to use glib/libpurple tools everywhere so it *should* be fine, but I'll test it on a windows VM later to make sure there isn't any weird platform specific behavior.

B1G-FUNGUS:
Actually, I've been having a little trouble figuring out what tools the makefile actually uses to compile an executable for windows. I tried both MXE and trying to use mingw natively on Windows, but I was having trouble with both. @EionRobb if you wouldn't mind compiling it or saying which tools you used then one of us can test this on Windows. Aside from that, I haven't encountered any more bugs, so I'm going to reopen this PR just in case someone else tests it and finds it merge-ready before I do.

Additionally, I've realized I've organized my commits in a silly way (eg; there are two separate commits which fix the same segfault), so if someone wants me to try reorganizing them just let me know.

alexolog:
> @alexolog thanks for the reminder. I made sure to use glib/libpurple tools everywhere so it _should_ be fine, but I'll test it on a windows VM later to make sure there isn't any weird platform specific behavior.

Thanks!  It would be great to have a DLL.

libdiscord.c
```diff
@@ -9730,6 +9735,276 @@ discord_cmd_get_history(PurpleConversation *conv, const gchar *cmd, gchar **args
...
+
+		// if there any error codes that are worth it to work around we
+		// can put them here
+		JsonNode *code_node = json_object_get_member(result, "code");
+		if (code_node != NULL) {
+			gint64 result_code = json_node_get_int(code_node);
+			gchar code_str[1024];
+			g_snprintf(code_str, 1024, "%ld", result_code);
+			const gchar *result_msg = json_object_get_string_member(result, "message");
+			purple_debug_error("discord", "xfer/http upload returned code: %ld and message:\n%s\n", result_code, result_msg);
```
EionRobb, line 9813: Ditto here with the `G_GINT64_FORMAT` macro :)

libdiscord.c
```diff
@@ -9730,6 +9735,276 @@ discord_cmd_get_history(PurpleConversation *conv, const gchar *cmd, gchar **args
...
+		return;
+	}
+	purple_xfer_set_size(xfer, file_len);
+
+	// though the gtk glib docs say otherwise, it appears that the contents
+	// are NOT the responsibility of the caller
+	gchar *contents = g_mapped_file_get_contents(file);
+
+	gboolean guessing;
+	mimetype = g_content_type_guess(fullpath, contents, file_len, &guessing);
```
EionRobb, line 9883: I'm getting a compiler error here on win32 that the second param needs to be a `guchar *` but a `gchar *` is being passed in. Does it just need an explicit `(guchar *)` cast to force the data in?

libdiscord.c
```diff
@@ -9730,6 +9735,276 @@ discord_cmd_get_history(PurpleConversation *conv, const gchar *cmd, gchar **args
...
+			return;
+		}
+
+		// if there any error codes that are worth it to work around we
+		// can put them here
+		JsonNode *code_node = json_object_get_member(result, "code");
+		if (code_node != NULL) {
+			gint64 result_code = json_node_get_int(code_node);
+			gchar code_str[1024];
+			g_snprintf(code_str, 1024, "%ld", result_code);
```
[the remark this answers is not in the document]
EionRobb: You could also use something similar to the `from_int()` helper: `g_strdup_printf("%" G_GINT64_FORMAT, id);`, which would need free'ing

libdiscord.c
EionRobb: Oh, if you've already got the `code_str` above, you could `%s` it instead

EionRobb:
Thanks for working on this @B1G-FUNGUS this is awesome :)

After making those comments, I figured it'd be better to just merge the PR as-is, since it works great on *nix and tweak up the Windows-specific bits

The build env is set up using the https://pidgin.im/development/building/2.x.y/windows/ instructions

EionRobb:
The dll at https://eion.robbmob.com/libdiscord.dll is now updated :)

B1G-FUNGUS:
This is a little late but thank you again for all the feedback. I probably should have found the pidgin windows build instructions but  I guess it slipped my mind.

Side note: I'm going to try and see if I can implement this for libpurple 3 as well, but I'm not sure if it's at a stage yet where file transfers are reasonably stable.

EionRobb:
@B1G-FUNGUS thank you for working through the feedback 🙂 

I wouldn't worry about purple3 for now as it's a bit of a moving target, and the rest of the plugin probably doesn't compile with it any more

PR #459, opened 2024-03-12
Startup rework

dlmarquis:
This reworks the startup so that long-standing bugs can be fixed.

Channel objects now include DMs and have a subtype ChannelData to help save memory space.

The mute function was removed because it has been removed from Discord's API.

Probably not ready for merge yet. I need to double check for any leaks or silly errors.

### DIFF
--- a/json_compat.h
+++ b/json_compat.h
@@ -36,6 +36,11 @@
 #define json_array_get_length(JSON_ARRAY) \
 	(JSON_ARRAY ? json_array_get_length(JSON_ARRAY) : 0)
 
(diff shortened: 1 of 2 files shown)


EionRobb:
I really like where this is heading! 

Not sure about the `—>d—>` name though, but naming stuff is hard and definitely not a blocker on merging this PR. 😄 

(Glad to see that the windows build tests are working with the PR, this is the first PR that's come through that's using them. Yay.) Will it be ok to make it work on windows too? Sing out if you need a hand with that 🙂

PR #482, opened 2026-03-19
Macos fixes

amarzot:
I was able to compile and use these modifications to chat on discord on my M1 mac

strangely these changes seem like they would be required regardless of platform, so unclear how they are working on others

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed memory management issue in Discord header generation that was preventing proper resource cleanup, leading to significantly improved application stability, reliability, and reduced potential for memory-related system issues.

* **New Features**
  * Added optional QR-code authentication capability, offering users a secure alternative login method that enhances accessibility and provides greater flexibility in authentication options.

### DIFF
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -1187,7 +1187,7 @@ discord_build_x_super_properties_header(DiscordAccount *da)
 	JsonObject *obj = discord_get_auth_properties(da);
 	gchar *json_str = json_object_to_string(obj);
 
-	g_object_unref(obj);
+	json_object_unref(obj);
 
 	cached_header = g_base64_encode((const guchar *)json_str, strlen(json_str));
 	g_free(json_str);
--- a/libdiscord.c
+++ b/libdiscord.c
@@ -10601,6 +10601,10 @@ typedef struct
 } PurplePluginProtocolInfoExt;
 
 
+#ifdef USE_QRCODE_AUTH
+#include <nss.h>
+#endif
+
 static void
 plugin_init(PurplePlugin *plugin)
 {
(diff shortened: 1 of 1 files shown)


EionRobb:
Thank you so much for the PR 🙂 

It's interesting that it needed this! Just to check, in the Help -> About window, what SSL plugin does it say is in use on your system? 

I'm wondering if NSS works on other platforms when Pidgin is already using NSS but fails if it's using openssl or gnutls (determined at runtime)

EionRobb:
Oh yeah looks like libpurple does that call for NSS already when init'ing NSS

And the docs say not to call it twice . https://www-archive.mozilla.org/projects/security/pki/nss/ref/ssl/sslfnc.html#1234224

Hmm. Tricky one. 
Maybe the plugin needs to check if NSS has already been loaded somehow and not call the init if it has already? Trouble is that libdiscord will be loaded before ssl-nss (as plugins are loaded in filesystem order) so we would have to do the init during the QR code generation process maybe?

amarzot:
> what SSL plugin does it say is in use on your system?

Found this in Help -> Plugin Information
```
Plugin Information
Discord
	Author: Eion Robb <eion@robbmob.com>
	Version: 0.9.2026.03.19.git.8cb88c4
	Website: https://github.com/EionRobb/purple-discord
	ID String: prpl-eionrobb-discord
	Loadable: Yes
	Loaded: Yes

...

GNUTLS
	Author: Christian Hammond <chipx86@gnupdate.org>
	Version: 2.14.14
	Website: https://pidgin.im/
	ID String: ssl-gnutls
	Loadable: Yes
	Loaded: Yes
...

SSL
	Author: Christian Hammond <chipx86@gnupdate.org>
	Version: 2.14.14
	Website: https://pidgin.im/
	ID String: core-ssl
	Loadable: Yes
	Loaded: Yes

...
```

Also unfortunately I think I can't use the plugin anymore as it trips something in discord and locks me out of my account for a while

EionRobb:
FYI, I cherry-picked through your commit for fixing the json object unref as that bits important and would hate for that to get lost in the other NSS changes in this PR

PR #491, opened 2026-09-04
Fix channel name being replaced by channel id in chat tabs

nothsaaaa:
When joining a server channel the tab title would initially show the channel name but then get overwritten with the channel's snowflake id because Pidgin re-runs purple_conversation_autoset_title() whenever a conversation update implies the display name may have changed (user list add/remove, chat left, account online/offline, account change) on GTK. For chats that aren't in the buddy list autoset_title() falls back to the conversation name, which is the numeric channel id.

Changes:
- Swap the chat_info field order so Name is the primary identifier instead of the id, which is what purple_chat_get_name() returns
- Set the conversation title to the channel name when joining and when channel info is fetched
- Add a conversation-updated handler that restores the channel name if Pidgin's autoset_title() clobbers it back to the id

I'm not primarily a C developer and understand there may be some errors in my code's safety that I haven't yet caught, but as for functionality, it does work. Further testing on other platforms and revision is requested. I'm on Fedora 43 KDE Plasma 6.1.

## Summary by CodeRabbit

- **Improvements**
  - Discord conversation titles are now maintained consistently instead of being replaced by channel IDs.
  - Conversation titles are set when opening a chat and refreshed when channel details or names are updated.
  - Chat creation details now display the conversation name before its ID for clearer identification.
  - Titles remain restored correctly when conversation updates are received.


EionRobb:
From memory, the reason we used id as the name instead of the name, was to handle duplicately named channels on different servers, eg trying to join general on one and general on another.  Does that seem to work ok after the change?

nothsaaaa:
Yes. The change affects only the displayed title and not the conversation identity. `discord_open_chat()` still creates the conversation with the numeric ID as the name and every lookup/join still uses the `discord_chat_hash(id)` which is unique per channel ID.

Unrelated to this PR, I did notice that with the default setting "Auto-create rooms on buddy list" *will* cause conflicts when there are multiple channels across multiple servers with the same name (eg "#general" in a server named "Tom's Diner", and "#general" in "John's Pub". If this is a concern, that might be worth looking at too. However, I believe the plugin just sets multiple channels with the same alias, then Pidgin then treats them as the same entries. As far as I can tell, the most recently added one wins in the buddy list. Worth looking at separately.

EionRobb:
Thanks for this PR!  I like the idea of hooking into the signal to fix the title when Pidgin ruins it and might steal that same idea for some other protocol plugins too :)

